Issue #3, opened 2020-06-17
Implement Skeleton for PID controller 


kammce:
Blah blah blah

Also add test skeleton

PR #8, opened 2020-06-25
Updated MPU6050 accelometer class

AndrewDEmerson:
Draft of brushed dc motor interface


Common/Mpu6050.hpp
```diff
@@ -0,0 +1,196 @@
...
+  {
+    /// Register address of the the first byte of the X axis
+    kXYZStartAddress = 0x3B,
+
+    /// Device ID register address
+    kWhoAmI = 0x75,
+
+    /// Device configuration starting address
+    kDataConfig = 0x1C,
+
```
kammce, line 28: Remove this extra line here
```suggestion
```

Common/Mpu6050.hpp
```diff
@@ -0,0 +1,196 @@
...
+  };
+
+  ///
+  ///
```
kammce: Say something here like

```suggestion
  /// MPU6050 Constructor
  ///
```

Common/Mpu6050.hpp
```diff
@@ -0,0 +1,196 @@
...
+      units::acceleration::standard_gravity_t full_scale = 2_SG,
+      uint8_t address                                    = 0x68)
+      : i2c_(i2c), kFullScale(full_scale), kAccelerometerAddress(address)
+  {
+  }
+
+  Returns<void> Initialize() override
+  {
+    Status status = i2c_.Initialize();
+    i2c_.Write(kAccelerometerAddress, { 0x6B, 0x0 });
```
kammce: Initialize should only setup peripherals. It shouldn't reach out to the device until `Enable` is called. Move this logic to `Enable`.

Common/Mpu6050.hpp
```diff
@@ -0,0 +1,196 @@
...
+
+  /// Will automatically set the device to 2g for full-scale. Use
+  /// `SetFullScaleRange()` to change it.
+  Returns<void> Enable() override
+  {
+    // Check that the device is valid before proceeding.
+    SJ2_RETURN_ON_ERROR(IsValidDevice());
+
+    // Put device into standby so we can configure the device.
+    // SJ2_RETURN_ON_ERROR(ActiveMode(false));
```
kammce: Remove if not needed. But make sure to verify in an example project that this actually works. It would be good to have example projects for each of our common drivers to test that they work properly.
```suggestion
```

kammce: But this is something that is nice to have when we attempt to bring this into SJSU-Dev2.

Common/Mpu6050.hpp
```diff
@@ -0,0 +1,196 @@
...
+
+    // Set device full-scale to the value supplied by the constructor.
+    SJ2_RETURN_ON_ERROR(SetFullScaleRange());
+
+    // Activate device to allow full-scale and configuration to take effect.
+    // SJ2_RETURN_ON_ERROR(ActiveMode(true));
```
kammce: Delete if not needed.
```suggestion
```

Common/Mpu6050.hpp
```diff
@@ -0,0 +1,196 @@
...
+    // 16 value.
+
+    int16_t x = static_cast<int16_t>(xyz_data[0] << 8 | xyz_data[1]);
+    int16_t y = static_cast<int16_t>(xyz_data[2] << 8 | xyz_data[3]);
+    int16_t z = static_cast<int16_t>(xyz_data[4] << 8 | xyz_data[5]);
+
+    //Convert the 16 bit value into a floating point value in SG
+    acceleration.x = kFullScale * (float)x / (INT16_MAX+1);
+    acceleration.y = kFullScale * (float)y / (INT16_MAX+1);
+    acceleration.z = kFullScale * (float)z / (INT16_MAX+1);
```
kammce: Tip #1 use `static_cast` for casting. Do NOT use C style casting. It is an unsafe operation.
Tip #2 refrain from using macros in C++ when C++ alternatives exists. For example

```C++
#include <limits>
// ...
int16_t max = std::numeric_limits<int16_t>::max();
```

Tip #3 use the map function here to map the input values to the output values. 

```suggestion
    constexpr int16_t kMax = std::numeric_limits<int16_t>::max();
    constexpr int16_t kMin = std::numeric_limits<int16_t>::min();
    
    float x_ratio = sjsu::Map(x, kMin, kMax, -1.0f, 1.0f);
    acceleration.x = kFullScale * x_ratio;
    // Do the same for the other axis
```

Common/Mpu6050.hpp
```diff
@@ -0,0 +1,196 @@
...
+    }
+    else if (gravity_scale == 16)
+    {
+      gravity_code = 0x03;
+    }
+    gravity_code = gravity_code << 3;
+    Status status =
+        i2c_.Write(kAccelerometerAddress,
+                   { Value(RegisterMap::kDataConfig), gravity_code });
+
```
kammce: Do you need to preserve any of the data in this field from before or is setting them to zero ok?

Common/brushed_dc.hpp
```diff
@@ -0,0 +1,25 @@
+#pragma once
```
kammce: Please put this in its own branch/commit. This should not be in this pull request.

kammce (changes requested):
Please remove the brushed DC motor interface form this PR and add it to a new one on its own.

The title is a bit confusing to read. Can you reword it.

Common/Mpu6050.hpp
AndrewDEmerson: ActiveMode function has been added

Common/Mpu6050.hpp
AndrewDEmerson: There is a high filter config in this address, so the function now reads the address and ORs the new values in

AndrewDEmerson:
@kammce can you review this?

Common/Mpu6050.hpp
```diff
@@ -0,0 +1,231 @@
...
+    return {};
+  }
+
+ private:
+  const I2c & i2c_;
+  const units::acceleration::standard_gravity_t kFullScale;
+  const uint8_t kAccelerometerAddress;
+};
+}  // namespace sjsu
```
kammce: Add new line to this file and then you can submit this.

kammce (approved):

kammce:
Just realized there is a bug in this code: https://github.com/SJSURoboticsTeam/urc-control_systems-2021/pull/8/files#diff-6ae17bf8528c6b1351515acad63fbf49R183

PR #18, opened 2020-07-28
Fixed Incorrect Bit Manipulation in MPU6050

### LINKED ISSUE #14: Bug in MPU6050
There needs to be a 0b in front of this bit mask at this line:

https://github.com/SJSURoboticsTeam/urc-control_systems-2021/pull/8/files#diff-6ae17bf8528c6b1351515acad63fbf49R183

AndrewDEmerson:
Fixed the issue with the bit masking by using the bit manipulation library. 
Fixed another issue where the MPU was put in sleep mode before setting the full scale range, causing it to not update the value.

Resolves SJSURoboticsTeam/urc-control_systems-2021#14

### DIFF
--- a/Common/Mpu6050.hpp
+++ b/Common/Mpu6050.hpp
@@ -179,16 +179,10 @@ class Mpu6050 : public Accelerometer
                        { Value(RegisterMap::kDataConfig) }, &controlRegister,
                        1);
     auto sleepMask = bit::MaskFromRange(6);
-    // A value of 1 in bit 6 of the control register means the MPU6050 should be
-    // in sleep mode
-    if (is_active)
-    {
-      controlRegister = bit::Clear(controlRegister, sleepMask);
-    }
-    else
-    {
-      controlRegister = bit::Set(controlRegister, sleepMask);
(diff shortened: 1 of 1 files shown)


Common/Mpu6050.hpp
```diff
@@ -180,11 +178,22 @@ class Mpu6050 : public Accelerometer
...
+    // A value of 1 in bit 6 of the control register means the MPU6050 should be
+    // in sleep mode
+    if (is_active)
+    {
+      controlRegister = bit::Clear(controlRegister, sleepMask);
+    }
+    else
+    {
+      controlRegister = bit::Set(controlRegister, sleepMask);
+    }
```
kammce: Rather than do a conditional here, you can simply insert the values. 
```suggestion
      // !is_active is required as the bit must be set to 0 in order to prevent it from sleeping.
      controlRegister = bit::Insert(controlRegister, !is_active, sleepMask);
```

kammce (changes requested):

kammce (approved):

PR #22, opened 2020-11-25
Add Multiplatform RMD X Tech Demo

kammce:
Add .gitignore file that excludes */build directories from the repo as
these folders usually contain binaries and other temp files that are not
needed in the main repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,14 @@
+**/*/build/
+**/*.gch
+**/*.gcno
+**/.gdb_history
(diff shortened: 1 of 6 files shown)


naterpotatoers (approved):

kammce:
@naterpotatoers its good practice to approve a PR but not submit it. Rationale here is that the author may actually want to update something that they spot. It makes sense to submit it for them, if they do not have submission privileges.

PR #23, opened 2020-12-03
Drive class scaffolder

naterpotatoers:
Initial working draft of the drive design document. Ignore spin.hpp, I had implemented it onto SJTwo board and would like to have it for reference.


kammce:
If this is a draft and not ready for review, you should convert this PR to a draft to make that apparent.

Drive/main.cpp
```diff
@@ -0,0 +1,25 @@
...
+  units::angle::degree_t angle;
+
+  RoverDriveSystems roverDriver;
+  roverDriver.Initialize();
+
+      while (true)
+  {
+    // {speed, angle, mode } = getMissionControlCommands()
+
+    if (mode == "drive")
```
kammce: Add curly braces

kammce: Convert into a switch case and use only single character (D = drive, S = spin, T = translation)

kammce:
```C++
enum class DriveModes
{
  kDrive = 'D',
  kSpin = 'S',
  kTranslation = 'T',
}
```

Drive/main.cpp
```diff
@@ -0,0 +1,25 @@
...
+  string mode;
+  float speed;
+  units::angle::degree_t angle;
+
+  RoverDriveSystems roverDriver;
+  roverDriver.Initialize();
+
+      while (true)
+  {
+    // {speed, angle, mode } = getMissionControlCommands()
```
kammce:
```C++
struct DriveCommand_t
{
  bool is_operational = false;
  char mode = 'S';
  float speed = 0;
  units::angle::degree_t angle = 0_deg;
};

DriveCommand_t GetMissionControlCommands() {
  return DriveCommand_t{};
}
```

Actually add in the function so that this can compile.

Drive/main.cpp
```diff
@@ -0,0 +1,25 @@
+#include "./RoverDriveSystems.hpp"
+
+int main()
+{
+  string mode;
+  float speed;
+  units::angle::degree_t angle;
```
kammce: Not sure why this is here?

Drive/RoverDriveSystems.hpp
```diff
@@ -0,0 +1,36 @@
...
+  sjsu::Servo steeringRightWheel;
+  sjsu::Servo steeringBackWheel;
+
+  // Create servo drive classes
+  sjsu::Servo driveLeftWheel;
+  sjsu::Servo driveRightWheel;
+  sjsu::Servo driveBackWheel;
+
+  // Three different driving modes
+  void handleDriveMode(float speed, units::angle::degree_t angle);          // Normal driving mode
```
kammce:
```suggestion
  void handleDriveMode();          // Normal driving mode
```

You don't need this if you are going to keep the values internally.

Drive/RoverDriveSystems.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+  // Create servo steering classes
+  sjsu::Servo steeringLeftWheel;
+  sjsu::Servo steeringRightWheel;
+  sjsu::Servo steeringBackWheel;
+
+  // Create servo drive classes
+  sjsu::Servo driveLeftWheel;
+  sjsu::Servo driveRightWheel;
+  sjsu::Servo driveBackWheel;
```
kammce:
```suggestion
```

Delete all of this. Lets not make assumptions about what we are using at this point.

kammce (changes requested):
This code should compile. Do not pseudo code it. Make code has all of the API calls and hooks required to run the rover, just without the implementation.

Drive/rover_drive_system.hpp
```diff
@@ -0,0 +1,124 @@
+#pragma once
+
+#include <string>
```
kammce: Why is this here?
```suggestion
```

Drive/rover_drive_system.hpp
```diff
@@ -0,0 +1,124 @@
...
+  enum class Mode : char
+  {
+    kDrive     = 'D',
+    kSpin      = 'S',
+    kTranslate = 'T'
+  };
+
+  RoverDriveSystem(sjsu::drive::Wheel left_wheel,
+                   sjsu::drive::Wheel right_wheel,
+                   sjsu::drive::Wheel back_wheel)
```
kammce: Should take these as a reference not by value.

Drive/spin.hpp
```diff
@@ -0,0 +1,70 @@
+/*
```
kammce: Lets put this file in its own directory rather than cluttering the drive directory with it. Maybe put it in the a `prototype/spin_testing.cpp` file with its own `main()` function call.

Drive/spin.hpp
```diff
@@ -0,0 +1,70 @@
...
+  Contains the spin mode handler. 
+  Early prototype - WIP
+*/
+
+#include "L1_Peripheral/lpc40xx/pwm.hpp"
+#include "L2_HAL/actuators/servo/servo.hpp"
+#include "utility/log.hpp"
+#include "utility/time.hpp"
+
+void spinMode(float speed, units::angle::degree_t angle)
```
kammce: And if you make this its own cpp file, you can just make this `int main()`

Drive/spin.hpp
```diff
@@ -0,0 +1,70 @@
...
+    sjsu::Delay(10s);
+
+    // Command the servo to go to 0 degrees as fast as the servo will allow.
+    servo1.SetAngle(0_deg);
+    servo2.SetAngle(0_deg);
+    servo3.SetAngle(0_deg);
+    sjsu::Delay(10s);
+  }
+  return 0;
+}
```
kammce: Make sure to add a new line at the end of files. (Google C++ standards thing)

Drive/wheel.hpp
```diff
@@ -0,0 +1,91 @@
+
```
kammce: NIT: remove extra line up here.

kammce (changes requested):

PR #24, opened 2021-01-06
Arm control prototype

AndrewDEmerson:
Initial prototype of the arm control class. Required to complete issue SJSURoboticsTeam/urc-central-2021#148.


Arm/RoverArmSystem.hpp
```diff
@@ -0,0 +1,58 @@
...
+  units::angle::degree_t elbow_pos;
+  units::angle::degree_t rotunda_pos;
+  units::angle::degree_t wrist_roll_pos;
+  units::angle::degree_t wrist_pitch_pos;
+
+ public:
+  RoverArmSystem();
+  bool Home();
+  bool Initialize();
+  bool Get_data();
```
kammce:
```suggestion
  bool GetData();
```

Arm/RoverArmSystem.hpp
```diff
@@ -0,0 +1,58 @@
...
+  bool Move_arm();
```
kammce: Define implementation here as well do not define outside like you are doing will cause linker errors if this file is included twice or more.

Arm/hand.hpp
```diff
@@ -0,0 +1,12 @@
+namespace sjsu::arm
+{
+class hand
```
kammce:
```suggestion
class Hand
```

Arm/RoverArmSystem.hpp
```diff
@@ -0,0 +1,58 @@
+#include "utility/units.hpp"
+#include "joint.hpp"
+#include "wrist_joint.hpp"
+
+namespace sjsu
```
kammce:
```suggestion
namespace sjsu::arm
```

Arm/joint.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+namespace sjsu::arm
+{
+// the joint class is used for the rotunda, elbow, and shoulder motors.
+class joint
+{
+ private:
+  // The minimum allowable angle the joint is able to turn to in normal
+  // operation.
+  units::angle::degree_t minimum_angle;
```
kammce:
```suggestion
  units::angle::degree_t minimum_angle = 0_deg;
```

Arm/joint.hpp
```diff
@@ -0,0 +1,46 @@
...
+  void set_zero_offset(units::angle::degree_t offset);
+  sjsu::Accelerometer::Acceleration_t get_accelerometer_data();
+};
+joint::joint()
+{
+  minimum_angle     = 0_deg;
+  maximum_angle     = 180_deg;
+  rest_angle        = 0_deg;
+  zero_offset_angle = 0_deg;
+}
```
kammce: This will always lead to a crash if you do not init mpu and motor member variables.

Arm/source/main.cpp
```diff
@@ -0,0 +1,13 @@
...
+
+int main()
+{
+  sjsu::RoverArmSystem armControl;
+  armControl.Initialize();
+  armControl.Home();
+  while (true)
+  {
+    armControl.Get_data();
+    armControl.Move_arm();
```
kammce: This NEEDS a delay. 
```suggestion
    armControl.Move_arm();
    sjsu::Delay(100ms);
```

kammce: Attempting to read data from the server as quickly as possible without any sort of delay will cause the server to be overloaded which is basically a DDOS on our own systems and will result in undefined behavior.

Arm/wrist_joint.hpp
```diff
@@ -0,0 +1,50 @@
...
+  wrist_joint()
+  {
+    pitch_minimum_angle     = 0_deg;
+    pitch_maximum_angle     = 40_deg;
+    pitch_rest_angle        = 20_deg;
+    roll_minimum_angle      = 0_deg;
+    roll_maximum_angle      = 180_deg;
+    roll_rest_angle         = 90_deg;
+    left_zero_offset_angle  = 0_deg;
+    right_zero_offset_angle = 0_deg;
```
kammce: Place these above so you don't need to worry about them in the constructor.

Arm/RoverArmSystem.hpp
```diff
@@ -0,0 +1,58 @@
...
+  bool Get_data();
+  bool Move_arm();
+};
+
+RoverArmSystem::RoverArmSystem()
+{
+  Rotunda  = sjsu::arm::joint();
+  Shoulder = sjsu::arm::joint();
+  Elbow    = sjsu::arm::joint();
+  Wrist    = sjsu::arm::wrist_joint();
```
kammce: How does `sjsu::arm::joint()` know which motor/Servo ID its supposed to operate with.

Arm/hand.hpp
```diff
@@ -0,0 +1,12 @@
...
+{
+  // The hand uses 5 'Actuonix PQ12-100-6-R' linear Actuators that use PWM
+  // signals to control their lengths.
+  //
+  // Need to confirm that that hand is
+  // controled by the same microcontrol as rest of arm, or is its own system.
```
kammce: Actually the hand has its own MCU and talks over UART you will need to command it that way rather than directly using PWM.

Arm/joint.hpp
```diff
@@ -0,0 +1,46 @@
...
+  // accelerometer attached to the joint that is used to home the arm
+  sjsu::Mpu6050 * mpu;
+
+ public:
+  joint();
+  joint(units::angle::degree_t min_angle,
+        units::angle::degree_t max_angle,
+        units::angle::degree_t rest_angle,
+        sjsu::RmdX joint_motor,
+        sjsu::Mpu6050 accelerometer);
```
kammce:
```suggestion
        sjsu::RmdX * joint_motor,
        sjsu::Mpu6050 * accelerometer);
```

Previously passing by value will ensure that the system crashes because the parameter objects will no longer be valid after the constructor finishes its work.

Arm/joint.hpp
```diff
@@ -0,0 +1,46 @@
...
+  void set_zero_offset(units::angle::degree_t offset);
+  sjsu::Accelerometer::Acceleration_t get_accelerometer_data();
```
kammce: maybe an init function?

Arm/joint.hpp
```diff
@@ -0,0 +1,46 @@
+#pragma once
+#include "utility/units.hpp"
+#include "L2_HAL/actuators/servo/rmd_x.hpp"
+#include "L2_HAL/sensors/movement/accelerometer/mpu6050.hpp"
+#include "L1_Peripheral/lpc40xx/can.hpp"
+
+namespace sjsu::arm
+{
+// the joint class is used for the rotunda, elbow, and shoulder motors.
+class joint
```
kammce:
```suggestion
class Joint
```

Arm/RoverArmSystem.hpp
kammce: This need information to know which is doing what.

kammce (changes requested):

Arm/RoverArmSystem.hpp
```diff
@@ -0,0 +1,77 @@
...
+  units::angle::degree_t elbow_pos;
+  units::angle::degree_t rotunda_pos;
+  units::angle::degree_t wrist_roll_pos;
+  units::angle::degree_t wrist_pitch_pos;
+
+ public:
+  RoverArmSystem(sjsu::arm::Joint rotunda,
+                 sjsu::arm::Joint shoulder,
+                 sjsu::arm::Joint elbow,
+                 sjsu::arm::WristJoint wrist)
```
kammce: If you want to take these as a parameter then take them as a reference. But you don't need to since all of this can be kept within the class. You just need a way to distinguish between each joint in some fashion in this class.

Arm/source/main.cpp
```diff
@@ -0,0 +1,47 @@
...
+  sjsu::RmdX rmd_left_wrist(can_network, 0x14B);
+  sjsu::RmdX rmd_right_wrist(can_network, 0x14C);
+
+  // Create an I2C object for the IMU's as well as all of the MPU objects for
+  // the arm.
+  sjsu::lpc40xx::I2c i2c(sjsu::lpc40xx::I2c::Bus::kI2c2);
+  sjsu::Mpu6050 mpu_rotunda(i2c, 0x68);
+  sjsu::Mpu6050 mpu_shoulder(i2c, 0x69);
+  sjsu::Mpu6050 mpu_elbow(i2c, 0x6A);
+  sjsu::Mpu6050 mpu_wrist(i2c, 0x6B);
```
kammce, line 32: Place a comment here stating that this may not work depending on if we use the address translator or if we use the chip select method. Will go over this in the arm meeting.

kammce (approved):
There are a lot of details that will need to be worked out but for now this looks decent.

naterpotatoers (approved):
Good starting point

PR #25, opened 2021-01-21
Initial drive modes implementation

naterpotatoers:


Drive/rover_drive_system.hpp
```diff
@@ -34,63 +44,130 @@ class RoverDriveSystem
...
+  bool GetMissionControlData()
   {
-    if (rover_is_operational_)
+    // GET /drive JSON value ?
+    if (true)  // change 'true' to something for localhost connection found
+    {
+      char response[] = {};
+      ParseMissionControlData(response);
+      Move(mission_control_data_.drive_mode,
+           mission_control_data_.rotation_angle, mission_control_data_.speed);
```
naterpotatoers: passing them by value since I want the drive mode handlers handling all the conversions to degree or rpm

Drive/rover_drive_system.hpp
```diff
@@ -34,63 +44,130 @@ class RoverDriveSystem
...
+      return true;
```
kammce: This probably shouldn't be within the GetData section of the code. Because its doing more than simply Getting data. It is now parsing it and also acting upon it.

Drive/rover_drive_system.hpp
```diff
@@ -34,63 +44,130 @@ class RoverDriveSystem
...
+  bool GetRoverData()
+  {
+    sjsu::LogInfo("left wheel speed: %f", left_wheel_.GetSpeed());
+    sjsu::LogInfo("left wheel position: %f", left_wheel_.GetPosition());
+    sjsu::LogInfo("right wheel speed: %f", right_wheel_.GetSpeed());
+    sjsu::LogInfo("right wheel position: %f", right_wheel_.GetPosition());
+    sjsu::LogInfo("back wheel speed: %f", back_wheel_.GetSpeed());
+    sjsu::LogInfo("back wheel position: %f", back_wheel_.GetPosition());
+    return true;
+  };
```
naterpotatoers: Will probably want to store this to a variable for updating server with current rover values

Drive/rover_drive_system.hpp
```diff
@@ -34,63 +44,130 @@ class RoverDriveSystem
...
+  {
+    if (mission_control_data_.is_operational)
     {
-      // checks if the mode is different than what is currently running
       if (mode != static_cast<char>(current_mode_))
       {
         SetMode(mode);
       }
-      HandleModeMovement(rotation_angle, speed);
+      HandleRoverMovement(rotation_angle, speed);
```
kammce: It might be a good idea to not attempt to move the rover after getting a new mode, but to request again what the new movement commands are. This is more of a safety I remember doing before.

Drive/rover_drive_system.hpp
```diff
@@ -99,76 +176,165 @@ class RoverDriveSystem
...
+  /// @param mode Three Modes: D (drive), S (spin), T (translation)
+  void SetMode(char mode = 'S')
+  {
+    switch (mode)
     {
-      HandleSpinMode();
-    }
-    if (current_mode_ == Mode::kTranslation)
+      case 'D':
+      case 'd':
```
kammce: We should limit this to just one. We shouldn't respect the usage of lower and upper case. The API should be consistent and concise.

Drive/rover_drive_system.hpp
```diff
@@ -99,76 +176,165 @@ class RoverDriveSystem
...
-  /// Handles the movement for their respective modes.
-  void HandleDriveMode();
-  void HandleSpinMode();
-  void HandleTranslationMode();
+  /// Aligns rover wheels all in the same direction, facing foward
+  void SetDriveMode()
+  {
+    units::angle::degree_t left_wheel_angle  = -45_deg;
+    units::angle::degree_t right_wheel_angle = -135_deg;
+    units::angle::degree_t back_wheel_angle  = 90_deg;
```
kammce: You can make these constexpr since they aren't changing.

Drive/wheel.hpp
```diff
@@ -23,68 +23,73 @@ class Wheel
...
   void SetSpeed(units::angular_velocity::revolutions_per_minute_t hub_speed)
   {
     if (hub_speed > kMaxHubSpeed)  // need use abs(hub_speed) ?
     {
       hub_speed = kMaxHubSpeed;
     }
+    if (hub_speed < kMaxNegativeHubSpeed)  // need use abs(hub_speed) ?
+    {
+      hub_speed = kMaxNegativeHubSpeed;
+    }
```
kammce: Replace this with std::clamp

Drive/wheel.hpp
```diff
@@ -23,68 +23,73 @@ class Wheel
...
     hub_speed_ = hub_speed;
     steer_motor_.SetSpeed(hub_speed);
```
kammce: Wouldn't you want to change the hub motor speed and not the steering motor speed? Where do you change the speed of the hub motor so that the wheels go forward. Wouldn't the steering motor just be used to change the angle of the motor?

Drive/rover_drive_system.hpp
```diff
@@ -99,76 +176,165 @@ class RoverDriveSystem
...
-    return true;
+    SetWheelSpeed(speed);
+  };
+
+  /// Handles translation mode. Adjusts all the wheels, keeping them parallel
+  void HandleTranslationMode(
+      units::angular_velocity::revolutions_per_minute_t speed,
+      units::angle::degree_t angle)
+  {
+    left_wheel_.SetPosition(angle);
```
kammce:
```suggestion
    left_wheel_.SetSteeringAngle(angle);
```

I feel like SetSteeringAngle make this more obvious what is happening here.

Drive/rover_drive_system.hpp
```diff
@@ -34,63 +44,130 @@ class RoverDriveSystem
...
-  ///  Sets the new mode for the rover. Will reduce rover movement speed to zero
-  ///  before changing wheel mode & tire alignment
-  void SetMode(char mode)
+  /// Sets all wheels to the speed provided. Wheel class handles max/min speeds
+  /// @param speed the new movement speed of the rover
+  void SetWheelSpeed(units::angular_velocity::revolutions_per_minute_t speed)
+  {
+    left_wheel_.SetSpeed(speed);
+    right_wheel_.SetSpeed(speed);
+    back_wheel_.SetSpeed(speed);
```
kammce: Would probably be a good idea to throw an linear interploation (aka EMA aka exponential moving average) somewhere in here to smooth out changes in speed. The acceleration for the motors is stupid high and will cause issues with the robot. Just an FYI. A bit more complex for this change though. I'd leave a comment saying you'll fix this later. A issue to go along with it too will be nice as well.

Drive/rover_drive_system.hpp
```diff
@@ -34,63 +44,130 @@ class RoverDriveSystem
...
+    sjsu::LogInfo("back wheel position: %f", back_wheel_.GetPosition());
+    return true;
+  };
+
+  /// Parses incoming data from mission control to command rover
+  /// @param response incoming JSON / string data
+  void ParseMissionControlData(char * response)
+  {
+    // parse data somehow using sscanf
+    sjsu::LogInfo("MISSION CONTROL RESPONSE:\n%s", response);
```
kammce, line 135: Add a TODO to actually parse the data here with an accompanying issue in github.

kammce (changes requested):
This is looking good. Good work Nate.

Drive/wheel.hpp
naterpotatoers: yeah lol good catch

Drive/rover_drive_system.hpp
naterpotatoers: didn't like constexpr so I assigned them as const

Drive/rover_drive_system.hpp
```diff
@@ -18,6 +19,16 @@ class RoverDriveSystem
...
   };
 
+  /// Mission controls possible input values
+  struct MissionControlData
+  {
+    bool is_operational;
+    char drive_mode;
+    float rotation_angle;
+    float speed;
+    char * http_GET_param;
```
naterpotatoers: ignore, I got ahead of myself will create new branch to implement soon

Drive/rover_drive_system.hpp
```diff
@@ -34,63 +45,142 @@ class RoverDriveSystem
...
+
+  // Handles GET /drive?parameters for rover drive system to mission control
+  /// @return returns true if connection is established from mission control
+  bool GetMissionControlData()
+  {
+    mission_control_data_.http_GET_param =
+        "?is_operational=%d&drive_mode=%c&left_wheel_speed=%f&right_wheel_"
+        "speed=%f&back_wheel_speed=%f",
+    mission_control_data_.is_operational, static_cast<char>(current_mode_),
+    left_wheel_.GetSpeed(), right_wheel_.GetSpeed(), back_wheel_.GetSpeed();
```
naterpotatoers: ignore, got ahead of myself

naterpotatoers (commented):
Resolved all original comments except for std::clamp in wheel class SetSpeed(). Created new issue for that.

Drive/rover_drive_system.hpp
naterpotatoers: okay done

Drive/rover_drive_system.hpp
naterpotatoers: Made it an if/else statement to resolve issue

Drive/wheel.hpp
naterpotatoers: opened an issue for this

Drive/rover_drive_system.hpp
naterpotatoers: created issue and added TODO. Issue #153

Drive/rover_drive_system.hpp
naterpotatoers: removed

Drive/rover_drive_system.hpp
kammce: Delete this code if it is not meant to be here.

Drive/rover_drive_system.hpp
kammce: Remove this if it is not meant to be here then. You shouldn't litter your code with a bunch of dead code. Each PR should be an implementation of a feature and no more. Dead code is not allowed.

Drive/rover_drive_system.hpp
```diff
@@ -120,6 +134,8 @@ class RoverDriveSystem
   /// @return true if rover is able to retrieve data from all the wheels
   bool GetRoverData()
```
kammce: Maybe call this `PrintRoverData()` because its not retrieving data from the motors or the rover. Its just printing it out. Which you won't see if you are operating the rover from afar without a connection to UART.

Drive/rover_drive_system.hpp
```diff
@@ -51,37 +52,48 @@ class RoverDriveSystem
...
+  /// speed adjusts the movement speed of the rover.
+  void Move()
   {
     if (mission_control_data_.is_operational)
     {
-      if (mode != static_cast<char>(current_mode_))
+      if (mission_control_data_.drive_mode != static_cast<char>(current_mode_))
       {
-        SetMode(mode);
+        SetMode(mission_control_data_.drive_mode);
```
kammce, line 83: Shouldn't you store the new `drive_mode` into `current_mode_` so you don't end up looping this forever?

kammce (changes requested):

Drive/rover_drive_system.hpp
naterpotatoers: SetMode() will update current_mode_ with the new drive mode

Drive/rover_drive_system.hpp
naterpotatoers: done

Drive/rover_drive_system.hpp
naterpotatoers: done

Drive/rover_drive_system.hpp
naterpotatoers: done - also changed it to a void function

naterpotatoers:
added in the requested changes

kammce (approved):

PR #26, opened 2021-02-01
Drive System Initial Implementation

### LINKED ISSUE #152: Parse Mission Control JSON Data
Implement arm and drive system JSON parser for incoming mission control data. 
Example JSON Data (string):
`{ "is_operational": 1, "drive_mode": "T", "speed": 80, "angle": 45} `

Extract the value such that we know
is_operational = true
drive_mode = T
speed = 80
angle = 45
Should be possible using [sscanf](https://www.youtube.com/watch?v=-7cSmcdMryo)

### LINKED ISSUE #154: Implement std::clamp for wheel class
SetSpeed() uses if statements to adjust speed input parameter speed if that value is outside the extrema bounds. Replace these if statements with [std::clamp ](https://en.cppreference.com/w/cpp/algorithm/clamp)

### LINKED ISSUE #155: Implement GET Request Handler for drive systems
Create GET request: {local-network-address}?key=value&key=value&...
key values refer to the current status of rover i.e. current mode, is_operational value, and wheel speeds.

### LINKED ISSUE #161: Unit Test Drive Class
Get unit tests started on ctrl sys. Aiming for > 70% code coverage

### LINKED ISSUE #162: Implement GETRequestConstructor() in Drive class
Need to implement GET request constructor for rover --> mission control.

naterpotatoers:
Adds in all the files for drive system. Also adds in the Esp class for making requests to mission control. Also includes unit tests for drive & wheel class.
- Resolves SJSURoboticsTeam/urc-central-2021#152 
- Resolves SJSURoboticsTeam/urc-central-2021#154
- Resolves SJSURoboticsTeam/urc-central-2021#155
- Resolves SJSURoboticsTeam/urc-central-2021#161
- Resolves SJSURoboticsTeam/urc-central-2021#162


Drive/rover_drive_system.hpp
```diff
@@ -22,10 +22,12 @@ class RoverDriveSystem
   /// Mission controls possible input values
   struct MissionControlData
   {
-    bool is_operational;
+    int is_operational;
```
naterpotatoers, line 20: Should behave the same way. I did this so that JSON parser wouldn't get mad at me for using %d to assign a boolean value.

kammce:
Don't put Closes #152 in the title. Put it in the commit message as well as the PR description message

Drive/rover_drive_system.hpp
```diff
@@ -127,17 +143,27 @@ class RoverDriveSystem
...
-    // hard coded for now
-    mission_control_data_.is_operational = true;
-    mission_control_data_.drive_mode     = 'S';
-    mission_control_data_.rotation_angle = 10.0f;
-    mission_control_data_.speed          = 30.0f;
+    const int expected_num_cmds = 4;
+    int parsed_num_cmds =
+        sscanf(mission_control_data_.response,
+               "{ \"is_operational\": %d, \"drive_mode\": "
+               "\"%c\", \"speed\": %f, \"angle\": %f}",
```
kammce: Rather than using all of these escape characters you can use raw strings.

Drive/rover_drive_system.hpp
kammce: Yeah this is fine. You'll need mission control to send you an into though, rather than a "true" or "false"

Drive/rover_drive_system.hpp
```diff
@@ -22,10 +22,12 @@ class RoverDriveSystem
   /// Mission controls possible input values
   struct MissionControlData
   {
-    bool is_operational;
+    int is_operational;
     char drive_mode;
     float rotation_angle;
     float speed;
+    char * response;
+    char * request;
```
kammce: This area looks dubious and dangerous. I feel like you are on your way to accessing invalid data with this here.

kammce (changes requested):

Drive/rover_drive_system.hpp
naterpotatoers: I didn't know this was a thing lol. My initial concern is whether the format specifiers will still work? I'll try it regardless

Drive/rover_drive_system.hpp
naterpotatoers: I ended up replacing with char var[300] but I'm not sure if that was right choice. I think std::string would make the most sense however that would mean parsing the GET response parsers would need to be redone... Curious what your thoughts are?

Drive/rover_drive_system.hpp
```diff
@@ -22,10 +23,12 @@ class RoverDriveSystem
   /// Mission controls possible input values
   struct MissionControlData
   {
-    bool is_operational;
+    int is_operational;
     char drive_mode;
     float rotation_angle;
     float speed;
+    char response_body[300];
+    std::string request_parameter;
```
naterpotatoers: Ideally would like to have these both as strings but response_body uses sscanf. I guess I could just alter it but wanted to know what @kammce thought first?

Drive/rover_drive_system.hpp
```diff
@@ -110,13 +108,48 @@ class RoverDriveSystem
...
+    mission_control_data_.request_parameter +=
+        "&left_wheel_angle=" + std::to_string(left_wheel_.GetPosition());
+    mission_control_data_.request_parameter +=
+        "&right_wheel_speed=" + std::to_string(right_wheel_.GetSpeed());
+    mission_control_data_.request_parameter +=
+        "&right_wheel_angle=" + std::to_string(right_wheel_.GetPosition());
+    mission_control_data_.request_parameter +=
+        "&back_wheel_speed=" + std::to_string(back_wheel_.GetSpeed());
+    mission_control_data_.request_parameter +=
+        "&back_wheel_angle=" + std::to_string(back_wheel_.GetPosition());
```
kammce: You can use `snprintf()` rather than this. I feel like having a format string and a list of parameters would be easier to read than this.

Drive/source/main.cpp
```diff
@@ -1,6 +1,6 @@
-#include "L1_Peripheral/lpc40xx/can.hpp"
-#include "L2_HAL/actuators/servo/rmd_x.hpp"
-#include "utility/units.hpp"
+#include "peripherals/lpc40xx/can.hpp"
+#include "devices/actuators/servo/rmd_x.hpp"
+#include "utility/math/units.hpp"
```
kammce: Make a PR just for adjusting/refactoring the code. Make another with just the changes for the drive json parser.

Drive/rover_drive_system.hpp
kammce: You can make them both character arrays if you like. I mentioned that you can use `snprintf` below, which is a better choice in terms of performance and eliminating memory fragmentation (does not dynamically allocate memory)

kammce (changes requested):

Drive/rover_drive_system.hpp
```diff
@@ -80,6 +84,10 @@ class RoverDriveSystem
                             mission_control_data_.speed);
       }
     }
+    else
+    {
+      sjsu::LogError("Rover is_operational is set to false!");
+    }
```
kammce, line 136:
```C++
void ExchangeMissionControlData() {
  GetRequest();
  ParseGetRequestBody();
  Move();
}
```

Just do this and have each function throw an exception if a part fails. Let main catch it and set the mode to a safe mode.

Drive/rover_drive_system.hpp
```diff
@@ -102,15 +110,40 @@ class RoverDriveSystem
...
+    int parameter_size = snprintf(
+        mission_control_data_.request_parameter, 300,
+        "?is_operational=%d&drive_mode=%c&battery=%d&left_wheel_speed=%f&"
+        "left_wheel_angle=%f&right_wheel_speed=%f&right_wheel_angle=%f&"
+        "back_wheel_speed=%f&back_wheel_angle=%f",
+        mission_control_data_.is_operational, static_cast<char>(current_mode_),
+        state_of_charge_, left_wheel_.GetSpeed(), left_wheel_.GetPosition(),
+        right_wheel_.GetSpeed(), right_wheel_.GetPosition(),
+        back_wheel_.GetSpeed(), back_wheel_.GetPosition());
+  };
```
kammce: Throw this into `GETRequest()` function

Drive/rover_drive_system.hpp
```diff
@@ -119,17 +152,31 @@ class RoverDriveSystem
...
+        mission_control_data_.response_body,
+        R"({"is_opertaional": %d, "drive_mode": "%c", "speed": %f, "angle": %f})",
+        &mission_control_data_.is_operational,
+        &mission_control_data_.drive_mode, &mission_control_data_.speed,
+        &mission_control_data_.rotation_angle);
+
+    if ((parsed_num_cmds == expected_num_cmds) &&
+        (mission_control_data_.drive_mode == 'D' ||
+         mission_control_data_.drive_mode == 'S' ||
+         mission_control_data_.drive_mode == 'T'))
```
kammce: Maybe throw into a simple function.

Drive/rover_drive_system.hpp
```diff
@@ -119,17 +152,31 @@ class RoverDriveSystem
...
+    {
+      return true;
+    }
+    else
+    {
+      sjsu::LogError("Error parsing response from Mission Control!");
+      return false;
```
kammce: Could thrown an exception here.

kammce (changes requested):

PR #29, opened 2021-03-09
WiFi Esp Class

### LINKED ISSUE #155: Implement GET Request Handler for drive systems

naterpotatoers:
Initial implementation of Wi-Fi module interface for rover.
Resolves SJSURoboticsTeam/urc-central-2021#155


Common/esp.hpp
```diff
@@ -0,0 +1,89 @@
...
+
+    socket_.Connect(sjsu::InternetSocket::Protocol::kTCP, host, kPort, 5s);
+    std::span request_payload(reinterpret_cast<const uint8_t *>(request.data()),
+                              request.size());
+    socket_.Write(request_payload, 5s);
+    std::array<uint8_t, 1024 * 2> response;
+    size_t read_back = socket_.Read(response, 10s);
+    // TODO: Return actual GET response
+    return "HTTP/1";
+  };
```
naterpotatoers: Struggling to return response as string. Never dealt with something like this before... @kammce

PR #34, opened 2021-07-28
Working Drive Parser

naterpotatoers:
Couldn't get the parser to work with type float so opted for using type int just for prototyping purposes. Although Maxim did bring up good point that type String could be a viable work around and upon thinking about it, it probably would work just don't have time today to implement / test it. If someone has the time to try that out that could be a permanent solution


Drive/rover_drive_system.hpp
```diff
@@ -60,11 +60,11 @@ class RoverDriveSystem
       // TODO - make these floats go to hundredths place (i.e. 0.00)?
       // Breaks unit test often since it never knows correct decimal value
       char reqParam[250];
-      snprintf(reqParam, 250,
+      snprintf(reqParam, 300,
+               "Vishnu-Adda/json-robo-test/"
```
kammce: Is the `Vishnu-Adda` part necessary?

kammce (approved):

Drive/rover_drive_system.hpp
naterpotatoers: For now it is. Once we figure out why localhost server can't be parsed by control sys code this will be removed. But for now the drive webserver is hosted at https://my-json-server.typicode.com/Vishnu-Adda/json-robo-test/drive. I'm going to look into that today a bit more

PR #35, opened 2021-08-26
Fixes ESP Problems

naterpotatoers:
Addresses the following problems as listed within the Slack channel for needed changes:

- [ ] Resilient to WiFi connection problems
- [ ] Resilient to server connection problems
- [ ] Resilient to slow connection problems
- [x] Remove needing to reconnect to WiFi hot spot every loop
- [x] Get connection rate to be continuous essentially


naterpotatoers:
Does not address WiFi connection problems / server problems. Will probably address those at later point. Issue created #SJSURoboticsTeam/urc-central-2021/issues/168 for those features. This contains the logic for just getting the homing going for now w/ new homing slip ring design and wifi problem fixes

PR #37, opened 2021-09-04
Updated SetHubSpeed using std::lerp

Maaz0070:


Drive/rover_drive_system.hpp
```diff
@@ -150,19 +150,104 @@ class RoverDriveSystem
...
+        increaseSpeed_leftWheel = false;
+      }
+      if(right_wheel_.RequestFeedbackFromMotor().GetFeedback().speed > speed)
+      {
+        increaseSpeed_rightWheel = false;
+      }
+      if(back_wheel_.RequestFeedbackFromMotor().GetFeedback().speed > speed)
+      {
+        increaseSpeed_backWheel = false;
+      }
```
kammce: We don't actually need to ask the motors for their current speed for this. We don't need true speed. The motors will handle getting to the speed a the rate they are programmed to. We just need to send them closer and closer steps to the goal. The won't be aligned but its also not necessary to be perfectly aligned with the output when it comes to lerp.

Drive/rover_drive_system.hpp
```diff
@@ -150,19 +150,104 @@ class RoverDriveSystem
...
+     
+
+      //lerp must continue as long as input speed(hub_speed) does not equal the output of the lerp function(lerp_speed) 
+       while((speed != std::clamp(units::angular_velocity::revolutions_per_minute_t{lerpSpeed_leftWheel}, kMaxNegSpeed, speed)) && (speed != std::clamp(units::angular_velocity::revolutions_per_minute_t{lerpSpeed_rightWheel}, kMaxNegSpeed, speed)) && (speed != std::clamp(units::angular_velocity::revolutions_per_minute_t{lerpSpeed_backWheel}, kMaxNegSpeed, speed)))
```
kammce: There shouldn't be any while loops in this algorithm.

Drive/rover_drive_system.hpp
```diff
@@ -150,19 +150,104 @@ class RoverDriveSystem
...
+        
+
+
+        sjsu::LogInfo("Set wheel speed to %f for all wheels", lerp_speed);
+
+        //make sure 100ms have passed before setting the next speed
+        while(Uptime() < next_same_time)
+        {
+          continue;
+        }
```
kammce: Yeah this won't work with the rest of the code as the system is locked here waiting for time to pass.

kammce (changes requested):

Drive/source/main.cpp
```diff
@@ -48,6 +48,9 @@ int main(void)
   sjsu::LogInfo("Initializing drive system...");
   drive_system.Initialize();
   esp.Initialize();
+  
+  drive_system.SetWheelSpeed(15_rpm);
+  drive_system.SetWheelSpeed(0_rpm);
```
naterpotatoers, line 53: Make sure to remove this and update with most recent main

Drive/rover_drive_system.hpp
```diff
@@ -150,19 +150,191 @@ class RoverDriveSystem
...
+      long double lerpSpeed_backWheel;
+
+      long double currentSpeed_leftWheel = static_cast<long double>(left_wheel_.GetSpeed());
+      long double currentSpeed_rightWheel = static_cast<long double>(right_wheel_.GetSpeed()); //GetSpeed);
+      long double currentSpeed_backWheel = static_cast<long double>(back_wheel_.GetSpeed());
+
+      long double new_speed = static_cast<long double>(speed);
+
+      //should be okay for almost all cases with 9 loops
+      for(int i = 0; i < 9; i++)
```
naterpotatoers: We shouldn't be looping anywhere within these member functions. This could potentially cause problems since this is a single threaded program. The only loop that should happen is within the main.cpp file.

Drive/rover_drive_system.hpp
```diff
@@ -150,19 +150,191 @@ class RoverDriveSystem
...
+      {
+        if((next_same_time % 10ms) == 0ms)
```
naterpotatoers: I think this logic might be somewhat problematic since %10 == 0ms could be unlikely. What if your timer is always 1ms too early or 1ms too late you'll never call this if statement. A possible solution might be to only lerp the wheels if at least 10ms has passed since the last time the wheels lerped?

Drive/rover_drive_system.hpp
```diff
@@ -150,19 +150,191 @@ class RoverDriveSystem
...
+        }
+
+       }//While
+  //  }//Try
     catch (const std::exception & e)
     {
       sjsu::LogError("Error setting wheels speed!");
       throw e;
     }
-  };
```
naterpotatoers: Should just delete this if your not planning on using it

naterpotatoers (changes requested):
Good progress so far. Highly recommend creating new branch with most recent master branch changes since there have been quite a few

naterpotatoers:
I think we are good to close this out ? Have the other PR #40 that resolves lerp

PR #39, opened 2021-10-26
moves ServerTimeout logic within esp class

naterpotatoers:
This should simplify the code found within arm and drive main files. This logic should have been contained within the class from the beginning imo.


naterpotatoers:
Does not work unfortunately. Closing this out for now

PR #40, opened 2021-11-13
Updated SetWheelSpeed in rover_drive_system using std::lerp

### LINKED ISSUE #171: Simultaneous wheel lerping
Original lerp mode was implemented within wheel class which meant rover would accelerate each wheel one by one. This would lead to future problems for the rover and is not correct way to implement lerp for our project.

Recreate lerp procedure from within drive_system file. End result should be a lerp handler that will quickly call each rover leg and set all the hub speeds at nearly the same time.

This should be non-blocking. Will require the use of some sort of timer

i.e.
```c++ 
// Main control loop
while()
{
   Action1();
  Action2();
   Action3();
  left_wheel.MoveWheel();
  right_wheel...
}

// rough example
MoveWheel(rpm target_speed) 
{
  if(!(Uptime() % 10)) // ~at least 10ms has passed, gives time for motor to move
  {
    motor_speed = std::lerp(previous_speed, target_speed, 0.01);
    motor(motor_speed);
    previous_speed = motor_speed;
  }
}

### PR DESCRIPTION
**Pull request by Maaz0070**
Updated SetWheelSpeed in rover_drive_system with std::lerp  
Resolves SJSURoboticsTeam/urc-central-2021/issues/171

### CODE REVIEW

File: Drive/wheel.hpp [hunk 1]
```diff
@@ -42,11 +42,19 @@ class Wheel
...
   {
-    hub_motor_.SetSpeed(hub_speed);
-    hub_speed_ = hub_speed;
-    // auto clampedHubSpeed = std::clamp(hub_speed, kMaxNegSpeed, kMaxPosSpeed);
-    // hub_motor_.SetSpeed(clampedHubSpeed);
-    // hub_speed_ = clampedHubSpeed;
+    try
+    {
+       sjsu::LogInfo("made it to sethubspeed()");
+
```
naterpotatoers: Probably good to remove this and clean up this function otherwise looks good

naterpotatoers (approved):

PR #55, opened 2022-01-03
Corey's branch

Coreyboy1820:


kammce:
Lets try and title the pull requests after what they are doing rather than the name of the contributor.

Common/rover_system.hpp
```diff
@@ -6,6 +6,10 @@ namespace sjsu::common
 class RoverSystem
 {
  public:
+  struct MissionControlData{
+    int is_operational;
+    int heartbeat_count;
+  };
```
naterpotatoers: @kammce Is this possible? This is to resolve SJSURoboticsTeam/urc-central-2021/issues/194. I've not dealt with a base class that has a struct within it that is inheritable...

Arm/joint.hpp
```diff
@@ -74,5 +74,11 @@ class Joint
   {
     return mpu.Read();
   }
+  void SetSpeed(units::angular_velocity::revolutions_per_minute_t speed)
+  {
+  speed_ = speed;
+  motor.SetSpeed(speed_);
+  }
```
naterpotatoers, line 81:
```suggestion
  void SetSpeed(int/double/float speed)
  {
  // Then cast speed variable into revolutions_per_minute_t...
  speed_ = speed;
  motor.SetSpeed(speed_);
  }
```

naterpotatoers: This removes all the casting needed to be done within rover_arm_system.cpp. Much cleaner imo

Arm/rover_arm_system.hpp
```diff
@@ -3,58 +3,199 @@
...
+    {
+      int pinky;
+      int ring;
+      int middle;
+      int pointer;
+      int thumb;
+    };
+    Fingers fingers;
+  };
+
```
naterpotatoers: Do we need speed variables? Is it important that arm moves at a certain speed? Or can this just be fixed to say 20rpm?

Arm/rover_arm_system.hpp
```diff
@@ -3,58 +3,199 @@
...
+        static_cast<float>(mc_data.shoulder_speed));
+    units::angle::degree_t shoulder_angle(
+        static_cast<float>(mc_data.shoulder_angle));
+    // elbow
+    units::angular_velocity::revolutions_per_minute_t elbow_speed(
+        static_cast<float>(mc_data.elbow_speed));
+    units::angle::degree_t elbow_angle(static_cast<float>(mc_data.elbow_angle));
+    // wrist
+    units::angular_velocity::revolutions_per_minute_t wrist_speed(
+        static_cast<float>(mc_data.wrist_speed));
```
naterpotatoers, line 140: all this casting could be avoided. This logic should be done in the implementation class imo. It clutters up this code base a bit too much

Arm/rover_arm_system.hpp
```diff
@@ -3,58 +3,199 @@
...
+    elbow_.SetPosition(elbow_angle);
+  }
+
+  void HomeArm()
+  {
+    // first getting the accelerometer readings/storing readings
+    Accelerometer::Acceleration_t rotunda_acceleration =
+        rotunda_.GetAccelerometerData();
+    Accelerometer::Acceleration_t shoulder_acceleration =
+        shoulder_.GetAccelerometerData();
```
naterpotatoers, line 109: Might be worth making a function or more likely an IMU / MPU class since this logic might be used over and over again

Arm/rover_arm_system.hpp
```diff
@@ -3,58 +3,199 @@
...
+    HomeShoulder(rotunda_acceleration, shoulder_acceleration);
+
+    Accelerometer::Acceleration_t elbow_acceleration =
+        elbow_.GetAccelerometerData();
+    HomeElbow(rotunda_acceleration, elbow_acceleration);
+
+    HomeWrist();
+  }
+  void HomeShoulder(Accelerometer::Acceleration_t rotunda_acceleration,
+                    Accelerometer::Acceleration_t shoulder_acceleration)
```
naterpotatoers, line 174: Might also be good idea to have Accelerometer::Acceleration_t member variables or another struct within this class for storing the acceleration values. Would remove the need to pass arguments which can get messy if not careful.

Arm/rover_arm_system.hpp
```diff
@@ -3,58 +3,199 @@
...
+      rotunda_acceleration_y = .0001;
+    }
+    if (shoulder_acceleration_x == 0.0)
+    {  // catches division by 0
+      shoulder_acceleration_x = .0001;
+    }
+    if (shoulder_acceleration_y == 0.0)
+    {  // catches division by 0
+      shoulder_acceleration_y = .0001;
+    }
```
naterpotatoers, line 198: This logic should be handled within a function helper or something. Doesn't seem right / hard to follow what exactly is going on.

Arm/rover_arm_system.hpp
```diff
@@ -3,58 +3,199 @@
...
+  void HomeShoulder(Accelerometer::Acceleration_t rotunda_acceleration,
+                    Accelerometer::Acceleration_t shoulder_acceleration)
+  {
+    double gravity                = 9.81;
+    double rotunda_acceleration_x = static_cast<double>(rotunda_acceleration.x);
+    double rotunda_acceleration_y = static_cast<double>(rotunda_acceleration.y);
+    double shoulder_acceleration_x =
+        static_cast<double>(shoulder_acceleration.x);
+    double shoulder_acceleration_y =
+        static_cast<double>(shoulder_acceleration.y);
```
naterpotatoers, line 182: This just feels wrong? How could we clean this up?

Arm/rover_arm_system.hpp
```diff
@@ -3,58 +3,199 @@
...
+    double rotunda_angle =
+        acos(gravity / rotunda_acceleration_y) *  // for angle
+        (rotunda_acceleration_x /
+         fabs(rotunda_acceleration_x));  // for direction
+    double shoulder_angle =
+        acos(gravity / shoulder_acceleration_y) *  // for angle
+        (shoulder_acceleration_x /
+         fabs(shoulder_acceleration_x));  // for direction
+    units::angle::degree_t home =
+        static_cast<units::angle::degree_t>(rotunda_angle + shoulder_angle);
```
naterpotatoers, line 208: We should build out functions for each of these calculation that way we can more easily test the logic and also cleanup code base.

Arm/rover_arm_system.hpp
```diff
@@ -3,58 +3,199 @@
...
+      rotunda_acceleration_y = .0001;
+    }
+    if (elbow_acceleration_x == 0.0)
+    {  // catches division by 0
+      elbow_acceleration_x = .0001;
+    }
+    if (elbow_acceleration_y == 0.0)
+    {  // catches division by 0
+      elbow_acceleration_y = .0001;
+    }
```
naterpotatoers, line 235: Duplicated logic from other function. Definitely needs to go / need a way to clean this up

Arm/rover_arm_system.hpp
```diff
@@ -3,58 +3,199 @@
...
+    double rotunda_angle =
+        acos(gravity / rotunda_acceleration_y) *  // for angle
+        (rotunda_acceleration_x /
+         fabs(rotunda_acceleration_x));  // for direction
+    double elbow_angle =
+        acos(gravity / elbow_acceleration_y) *  // for angle
+        (elbow_acceleration_x / fabs(elbow_acceleration_x)) *
+        (elbow_acceleration_y / fabs(elbow_acceleration_y));  // for direction
+    units::angle::degree_t home =
+        static_cast<units::angle::degree_t>(rotunda_angle + elbow_angle);
```
naterpotatoers, line 245: Same as above - We should build out functions for each of these calculation that way we can more easily test the logic and also cleanup code base.

Arm/rover_arm_system.hpp
```diff
@@ -76,5 +217,14 @@ class RoverArmSystem : public sjsu::common::RoverSystem
   void MoveElbow(){};
 
   void MoveWrist(){};
+
+  MissionControlData mc_data;
```
naterpotatoers, line 265:
```suggestion
  MissionControlData mc_data_;
```
I see drive system also has this incorrect @acarrou @ryerlic @Maaz0070

naterpotatoers (changes requested):

Arm/joint.hpp
naterpotatoers: Same goes for SetPosition() function

Common/rover_system.hpp
kammce: This should not be in this change. A new branch and PR should be made for this. But yes, to answer your question, this is very much possible.

kammce: See this: https://godbolt.org/z/Wb7z5KGMj

Arm/rover_arm_system.hpp
naterpotatoers: Also need to be careful with checking equality with double/floats. I had problems when running unit tests where I was expecting 0.00 and got 0.000000000000000001 or something like that that broke everything. This was more problematic when passing double between mission control so this might be fine, need to test it

naterpotatoers (approved):
Will need stories to implement suggested changes. For now this is good.

Arm/rover_arm_system.hpp
Coreyboy1820: The best way to clean this up would be to change the return type I get from the GetAcclerometerData function from the acceleration_t to a double so I can avoid this conversion. The rotunda_acceleration is a joint so I would ultimately have to change that class.

PR #56, opened 2022-01-03
Simpler unit conversion chage; try and catch deletions in drive class

### LINKED ISSUE #192: Remove Try/Catch Blocks
Only Try/Catch block should exist at top level in main.cpp. Lower level logic should handle potential errors or throw them. The first being ideal

acarrou:
Resolves SJSURoboticsTeam/urc-central-2021/issues/192

### DIFF
--- a/Drive/rover_drive_system.hpp
+++ b/Drive/rover_drive_system.hpp
@@ -25,6 +25,9 @@ const char message_format[] =
 class RoverDriveSystem : public sjsu::common::RoverSystem
 {
  public:
+  struct ParseError
+  {
+  };
   struct MissionControlData
   {
     int heartbeat_count;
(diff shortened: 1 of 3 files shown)


Drive/rover_drive_system.hpp
```diff
@@ -42,47 +45,31 @@ class RoverDriveSystem : public sjsu::common::RoverSystem
...
-    }
-    catch (const std::exception & e)
-    {
-      sjsu::LogError("Error initializing!");
-      throw e;
-    }
+    left_wheel_.Initialize();
+    right_wheel_.Initialize();
+    back_wheel_.Initialize();
+    HomeWheels();
```
naterpotatoers: I thought we found this line to be a bug. Should be 
```suggestion
    SetSpinMode();
```

naterpotatoers: That and we should probably initialize all the variables within the classes to prevent errors like this from reoccurring but that can be another issue / pr

Drive/rover_drive_system.hpp
```diff
@@ -275,124 +230,77 @@ class RoverDriveSystem : public sjsu::common::RoverSystem
...
+      return -0.378 + -1.79 * abs(angle) + 0.0366 * pow(abs(angle), 2) +
+             -3.24E-04 * pow(abs(angle), 3);
     }
     else
     {
-      return static_cast<double>(-(-0.378 + -1.79 * abs(angle) +
-                                   0.0366 * pow(abs(angle), 2) +
-                                   -3.24E-04 * pow(abs(angle), 3)));
+      return -(-0.378 + -1.79 * abs(angle) + 0.0366 * pow(abs(angle), 2) +
+               -3.24E-04 * pow(abs(angle), 3));
```
naterpotatoers, line 303: Does the logic still work? Will the return value just get truncated e.g. the result is 24.5918 --> return 24? Might want to write some tests to verify this.

Drive/wheel.hpp
```diff
@@ -43,17 +43,20 @@ class Wheel
...
 
   /// Adjusts the steer motor by the provided rotation angle/degree.
   /// @param rotation_angle positive angle (turn right), negative angle (left)
-  void SetSteeringAngle(units::angle::degree_t rotation_angle)
+  void SetSteeringAngle(int rotation_angle)
   {
-    steer_angle_ = rotation_angle + homing_offset_angle_;
+    // units::angle::degree_t steer_angle(rotation_angle + homing_offset_angle_);
+    units::angle::degree_t angle_into_deg(rotation_angle);
+    steer_angle_ = angle_into_deg;
```
naterpotatoers, line 59: @Coreyboy1820 example of passing in arguments as int and then doing the conversion within this function to simplify rover_drive_system.cpp file

naterpotatoers (approved):
Nice work

Drive/rover_drive_system.hpp
acarrou: yes I verified this

> Does the logic still work? Will the return value just get truncated e.g. the result is 24.5918 --> return 24? Might want to write some tests to verify this.

Drive/rover_drive_system.hpp
acarrou: > That and we should probably initialize all the variables within the classes to prevent errors like this from reoccurring but that can be another issue / pr

will do this in another pr then

PR #59, opened 2022-01-05
Drive Clean Up V2

naterpotatoers:
Went through and cleaned up wheel.hpp and drive system a bit more
- Changes tons of wheel.hpp variables to double / float removing units library stuff for the most part
- Renames functions for better clarity
- Adds SlipRingHomeWheel() in addition to HomeWheel(). HomeWheel() is only for testing purposes without slip ring. Since it's for testing I thought removing the spinning logic would remove unnecessary wait times

### DIFF
--- a/Drive/rover_drive_system.hpp
+++ b/Drive/rover_drive_system.hpp
@@ -86,28 +86,25 @@ class RoverDriveSystem : public sjsu::common::RoverSystem
     }
   };
 
-  /// Verifies that mission control is sending fresh commands
-  bool isSyncedWithMissionControl()
-  {
-    if (mc_data_.heartbeat_count != heartbeat_count_)
-    {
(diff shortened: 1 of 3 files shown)


naterpotatoers:
@acarrou can you verify that this logic works properly on your sj2 board with web server? Might've broken some stuff

acarrou:
> @acarrou can you verify that this logic works properly on your sj2 board with web server? Might've broken some stuff

Just tested. Works on the board, the only thing is the first home wheel button isn't working like it used to. Not sure why

kammce:
Ping me when you know this works.

naterpotatoers:
@kammce it works now 😅 @acarrou ran it on his machine and we fixed a few other issues at the same time. It wasn't lerping when going from high speed to low speed

PR #61, opened 2022-01-05
Fixed Units For Cleaner Code

Coreyboy1820:
Everything in this PR relates to cleaning up code via unit conversions:
-Changed all parameters of functions to doubles for cleaner code.
-Put all conversions into joint class instead of Rover_Arm_system.
-Moved private variables in the Joint class to the bottom.
-Added a structure of the accelerometer values as doubles instead of meters_per_second_squared.
-The GetAccelerometerValues now returns the structure Acceleration which consists of doubles to lessen the conversions and clean up code.

### DIFF
--- a/Arm/joint.hpp
+++ b/Arm/joint.hpp
@@ -8,24 +8,13 @@ namespace sjsu::arm
 // the Joint class is used for the rotunda, elbow, and shoulder motors.
 class Joint
 {
- private:
-  // The minimum allowable angle the joint is able to turn to in normal
-  // operation.
-  units::angle::degree_t minimum_angle = 0_deg;
-  // The maximum allowable angle the joint is able to turn to in normal
-  // operation.
-  units::angle::degree_t maximum_angle = 180_deg;
-  // The angle the joint will move to when is not operational
-  units::angle::degree_t rest_angle = 0_deg;
-  // The angle between the motor's zero position and the actual homed zero
-  // positon.
-  units::angle::degree_t zero_offset_angle = 0_deg;
-  // Motor object that controls the joint
-  sjsu::RmdX & motor;
-  // accelerometer attached to the joint that is used to home the arm
-  sjsu::Mpu6050 & mpu;
-
- public:
+  public:
+  struct Acceleration
+  {
+    double x;
+    double y;
+    double z;
+  };
   Joint(sjsu::RmdX & joint_motor, sjsu::Mpu6050 & accelerometer)
       : motor(joint_motor), mpu(accelerometer)
   {
@@ -53,9 +42,10 @@ class Joint
   }
 
   /// Move the motor to the (calibrated) angle desired.
-  void SetPosition(units::angle::degree_t angle)
+  void SetPosition(double angle)
   {
-    units::angle::degree_t calibrated_angle = angle - zero_offset_angle;
+    units::angle::degree_t angle_to_degrees(angle);
+    units::angle::degree_t calibrated_angle = angle_to_degrees - zero_offset_angle;
     calibrated_angle                        = units::math::min(
         units::math::max(calibrated_angle, minimum_angle), maximum_angle);
     sjsu::LogInfo("%f", calibrated_angle.to<double>());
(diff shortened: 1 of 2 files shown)


Arm/joint.hpp
```diff
@@ -8,24 +8,13 @@ namespace sjsu::arm
...
-  sjsu::Mpu6050 & mpu;
-
- public:
+  public:
+  struct Acceleration
+  {
+    double x;
+    double y;
+    double z;
+  };
```
naterpotatoers, line 17: nice

Arm/joint.hpp
```diff
@@ -53,9 +42,10 @@ class Joint
...
 
   /// Move the motor to the (calibrated) angle desired.
-  void SetPosition(units::angle::degree_t angle)
+  void SetPosition(double angle)
   {
-    units::angle::degree_t calibrated_angle = angle - zero_offset_angle;
+    units::angle::degree_t angle_to_degrees(angle);
+    units::angle::degree_t calibrated_angle = angle_to_degrees - zero_offset_angle;
     calibrated_angle                        = units::math::min(
         units::math::max(calibrated_angle, minimum_angle), maximum_angle);
```
naterpotatoers, line 50: is this logic for making sure joint doesn't go beyond max or min angles?
could be rewritten using std::clamp e.g.:
```suggestion
    calibrated_angle = clamp(calibrated_angle, minimum_angle, maximum_angle); 
```

naterpotatoers: oh but it might freak out about the types since clamp might not work nicely with units library

Arm/joint.hpp
```diff
@@ -64,21 +54,48 @@ class Joint
...
+  units::angle::degree_t rest_angle = 0_deg;
+  // The angle between the motor's zero position and the actual homed zero
+  // positon.
+  units::angle::degree_t zero_offset_angle = 0_deg;
+  // Motor object that controls the joint
+  sjsu::RmdX & motor;
+  // accelerometer attached to the joint that is used to home the arm
+  sjsu::Mpu6050 & mpu;
+  // speed of the joint motor
+  units::angular_velocity::revolutions_per_minute_t speed_;
```
naterpotatoers, line 99: I would remove all the comments, these variables are fairly self explanatory. The comments don't add any real value other than make us glaze over them without much consideration.

Arm/rover_arm_system.hpp
```diff
@@ -17,22 +17,22 @@ class RoverArmSystem : public sjsu::common::RoverSystem
...
-      int pinky;
-      int ring;
-      int middle;
-      int pointer;
-      int thumb;
+      double pinky;
+      double ring;
+      double middle;
+      double pointer;
+      double thumb;
```
naterpotatoers, line 35: This might be problematic since we've had troubles in the past parsing doubles from the GET request response but it is fine for now. Maybe when @tbejaoui1 or @acarrou start work on arm --> mission control logic we will find out and can change later

Arm/rover_arm_system.hpp
```diff
@@ -121,129 +121,100 @@ class RoverArmSystem : public sjsu::common::RoverSystem
...
+    Joint::Acceleration rotunda_acceleration =
         rotunda_.GetAccelerometerData();
-    Accelerometer::Acceleration_t shoulder_acceleration =
+    Joint::Acceleration shoulder_acceleration =
         shoulder_.GetAccelerometerData();
-    // then homing the different parts of the arm
     HomeShoulder(rotunda_acceleration, shoulder_acceleration);
 
-    Accelerometer::Acceleration_t elbow_acceleration =
+    Joint::Acceleration elbow_acceleration =
```
naterpotatoers, line 146: Might be worth making these local member variables for holding acceleration data that way we could possibly break it up later or run some checks on them or something. Might be a bit silly though, who knows

Arm/rover_arm_system.hpp
```diff
@@ -121,129 +121,100 @@ class RoverArmSystem : public sjsu::common::RoverSystem
...
+        (shoulder_acceleration.x /
+         fabs(shoulder_acceleration.x));  // for direction
+    double home = rotunda_angle + shoulder_angle;
+    MoveShoulder(10.0, home);
   }
 
-  void HomeElbow(Accelerometer::Acceleration_t rotunda_acceleration,
-                 Accelerometer::Acceleration_t elbow_acceleration)
+  void HomeElbow(Joint::Acceleration rotunda_acceleration,
+                 Joint::Acceleration elbow_acceleration)
```
naterpotatoers, line 189: Nitpick - might be nice to have this under the same namespace somehow. Seeing Joint:: everywhere doesn't look nice

Arm/rover_arm_system.hpp
```diff
@@ -121,129 +121,100 @@ class RoverArmSystem : public sjsu::common::RoverSystem
...
-      shoulder_acceleration_x = .0001;
+      shoulder_acceleration.x = .0001;
     }
-    if (shoulder_acceleration_y == 0.0)
+    if (shoulder_acceleration.y == 0.0)
     {  // catches division by 0
-      shoulder_acceleration_y = .0001;
+      shoulder_acceleration.y = .0001;
     }
+//TODO: clean up calculations into helper functions to make it more legible
```
naterpotatoers, line 175: Created new issue for this SJSURoboticsTeam/urc-central-2021/issues/200

naterpotatoers (approved):
Approving for now, good place for others to help out with now

Arm/joint.hpp
Coreyboy1820: I might have to redo quite a bit of the joint class, I didn't write that function so I'm trying to understand/edit what I can. I have no idea what the last guy was attempting to do there.

Arm/rover_arm_system.hpp
Coreyboy1820: Got it, I just wanted to stay consistent with the types I was using, I'll change those in the future when needed.

Arm/rover_arm_system.hpp
Coreyboy1820: I'll see what I can do.

naterpotatoers:
Feel free to merge this now and make the changes with other pull requests. I generally prefer that approach at the beginning

PR #62, opened 2022-01-06
Adds unit tests for drive

naterpotatoers:
Had trouble setting up mock for GPIO but I found a unit test from dev2 as an example. Does this look correct somewhat? It runs but just doesn't look pretty and might have unnecessary stuff.

### DIFF
--- a/Drive/test/wheel_test.cpp
+++ b/Drive/test/wheel_test.cpp
@@ -1,15 +1,11 @@
 #include "testing/testing_frameworks.hpp"
-#include "peripherals/lpc40xx/can.hpp"
 #include "devices/actuators/servo/rmd_x.hpp"
-#include "utility/log.hpp"
-#include "utility/math/units.hpp"
 
-#include "rover_drive_system.hpp"
 #include "wheel.hpp"
 
 namespace sjsu
 {
-TEST_CASE("Testing Wheel")
+TEST_CASE("Wheel testing")
 {
   Mock<Can> mock_can;
   Fake(Method(mock_can, Can::ModuleInitialize));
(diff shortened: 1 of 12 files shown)


Drive/test/wheel_test.cpp
```diff
@@ -1,75 +1,110 @@
...
+  SECTION("Spying on Wheel")
   {
-    wheel.HomeWheel();
-    CHECK(wheel.homing_offset_angle_ == 0_deg);
+    // TODO: Get spy logic working correctly - idk why it's not working...
+    Mock<drive::Wheel> spy(wheel);
+    When(Method(spy, GetSteerAngle)).AlwaysReturn(10);
+    drive::Wheel & test = spy.get();
+    CHECK(test.GetSteerAngle() == 10);
   }
```
naterpotatoers: @kammce I've been trying to figure out how spying works in FakeIt but for whatever reason this does not work. It returns 0 as opposed to AlwaysReturn() = 10. Used [this ](https://github.com/eranpeer/FakeIt/wiki/Quickstart#spying)as a guide with no luck. Have any ideas?

kammce: FakeIt can only work on virtual functions. In Java all functions are virtual functions which is why you can override any function you like. With C++ a normal function is hard coded and you cannot get the application to redirect to another function address so easily at runtime (should be impossible by principle). So if you wan this mockable by FakeIt, you'll need to make the function virtual.

This was actually one of my inspirations for why SJSU-Dev2 is interface heavy and why almost every driver has to be an interface. So that every driver can have its implementation swapped out on the fly in the unit testing library. 

For ease of use now, and not refactoring everything for better unit testing capability, I'd say, just make the methods you want to test virtual and we will optimize this later. :)

naterpotatoers: ohhh okay makes sense now, thanks will do

PR #72, opened 2022-01-07
Enabling Unit Testing + updated readability

Coreyboy1820:
-Created the getposition function and the position variable for the joint class to allow for unit testing.
-Created a new "VectorAddition" function to make reading the homing easier.

### DIFF
--- a/Arm/rover_arm_system.hpp
+++ b/Arm/rover_arm_system.hpp
@@ -136,12 +136,12 @@ class RoverArmSystem : public sjsu::common::RoverSystem
     // angle needed
 
     double acceleration_x =
-        accelerations_.rotunda.x +
-        accelerations_.shoulder.x;  // might need compliment value of shoulder
-    double acceleration_y =
-        accelerations_.rotunda.y + accelerations_.shoulder.y;
+        VectorAddition(accelerations_.rotunda.x, accelerations_.shoulder.x);  // might need compliment value of shoulder
+    double acceleration_y = 
+        VectorAddition(accelerations_.rotunda.y, accelerations_.shoulder.y);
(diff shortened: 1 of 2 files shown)


Arm/rover_arm_system.hpp
```diff
@@ -210,6 +211,19 @@ class RoverArmSystem : public sjsu::common::RoverSystem
...
+  {
+    if (acceleration == 0)
+    {
+      acceleration = 0.0001;
+    }
+  }
+  double VectorAddition(double left, double right)
+  {
+    return(left + right);
+  }
```
naterpotatoers, line 224: Not sure if this was necessary? Are there going to be edge cases that this will account for in the future?
Because it might actually be cleaner just doing the addition there e.g. 
`accel = accel_x + accel_y;`
as opposed to
`accel = VectorAddition(accel_x, accel_y);`

Arm/rover_arm_system.hpp
```diff
@@ -210,6 +211,19 @@ class RoverArmSystem : public sjsu::common::RoverSystem
   }
 
  private:
+  void VerifyNonZeroes(double & acceleration)
```
naterpotatoers, line 214:
```suggestion
  void IncrementIfZero(double & acceleration)
```
Think this is a bit more correct. Verify makes me think it's a boolean funciton that return true/false and wouldn't alter the state of the value.

naterpotatoers (approved):
few nitpick comments otherwise looks good

Arm/rover_arm_system.hpp
Coreyboy1820: Got it, wasn't sure if that was a good idea or not.

PR #76, opened 2022-01-08
Arm server config requesting review

tbejaoui1:
hey 
Added in the GetParameters stuff and added in an int variable called: 'state_of_charge' for the battery under the private section for the script titled: 'rover_arm_system.hpp'.

Thnx.

### DIFF
--- a/Arm/rover_arm_system.hpp
+++ b/Arm/rover_arm_system.hpp
@@ -7,6 +7,23 @@
 
 namespace sjsu::arm
 {
+const char response_body_format[] =
+  "\r\n\r\n{\n"
+  "  \"heartbeat_count\": %d,\n"
+  "  \"is_operational\": %d\n",
+  "  \"arm_speed\": %d,\n"
+  "  \"rotunda_angle\": %d,\n"
+  "  \"shoulder_angle\": %d,\n"
+  "  \"elbow_angle\": %d,\n"
+  "  \"wrist_roll\": %d,\n"
+  "  \"wrist_pitch\": %d,\n"
+  "  \"pinky_angle\": %d,\n"
+  "  \"ring_angle\": %d,\n"
+  "  \"middle_angle\": %d,\n"
+  "  \"pointer_angle\": %d,\n"
+  "  \"thumb_angle\": %d\n"
+  "}";
+
 class RoverArmSystem : public sjsu::common::RoverSystem
 {
  public:
--- a/Arm/rover_arm_system.hpp
+++ b/Arm/rover_arm_system.hpp
@@ -8,32 +8,37 @@
 namespace sjsu::arm
 {
 const char response_body_format[] =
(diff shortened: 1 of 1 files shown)


Arm/rover_arm_system.hpp
```diff
@@ -60,23 +82,70 @@ class RoverArmSystem : public sjsu::common::RoverSystem
...
+      elbow_angle_.GetPosition(),wrist_roll_.GetSpeed(),wrist_pitch_.GetSpeed(),
+      pinky_angle_.GetPosition(),ring_angle_.GetPosition(),middle_angle_.GetPosition(),
+      pointer_angle_.GetPosition(),thumb_angle_.GetPosition());
+
+    return request_parameter;
+
   };
 
-  std::string ParseJSONResponse()
+  void ParseJSONResponse(std::string & response)
```
naterpotatoers, line 105: is there a reason we are passing by reference? I guess it's not a big deal just wondering why

Arm/rover_arm_system.hpp
```diff
@@ -60,23 +82,70 @@ class RoverArmSystem : public sjsu::common::RoverSystem
...
+      heartbeat_count_ = 0;
+      return false;
+    }
+    return true;
+  }
+
+  void IncrementHeartbeatCount()
+  {
+    heartbeat_count_++;
+  }
```
naterpotatoers, line 150: Now that we have duplicated code for both arm and drive it might be a good idea to make these functions apart of the base class rover_system.hpp. That way if they ever get modified we only have to edit the code in one place. We can do this in a new pull request though.

Arm/rover_arm_system.hpp
```diff
@@ -216,7 +285,9 @@ class RoverArmSystem : public sjsu::common::RoverSystem
   sjsu::arm::WristJoint & wrist_;
   MissionControlData mc_data_;
   Acceleration accelerations_;
-  int heartbeat_count_                    = 0;
+  int heartbeat_count_   = 0;
+  int state_of_charge_ = 90;
+  const int kExpectedArguments  = 5;
```
naterpotatoers, line 292: Same with these variables, it might make sense to move them into rover_system.hpp at a later point to remove code duplication

naterpotatoers (approved):
nicely done. Feel free to merge in
Created a new github issue SJSURoboticsTeam/urc-central-2021/issues/206 for moving the duplicated logic to base class in rover_system.hpp

Arm/rover_arm_system.hpp
acarrou: My guess is that we don't want to copy it to the arguments since this is very sensitive as we've seen before

Arm/rover_arm_system.hpp
acarrou: Totally agree, I think that's something we should discuss today in our meeting.

PR #78, opened 2022-01-08
Min max arm

Andrew-Ingerman:
I'm getting a lot of errors saying that the variables aren't declared in the scope. But my changes seem to be compiling.


Arm/test/joint_test.cpp
```diff
@@ -52,24 +52,45 @@ TEST_CASE("Drive system testing")
...
+
+    joint.SetPosition(200);
+    CHECK(joint.GetPosition() == maxPosition);
+  }
+
   SECTION("Should set motor speed to 5, 10 , 45, 100")
+  {
+    joint.SetSpeed(5);
+    CHECK(joint.GetSpeed() == 2.5);
+  }
```
naterpotatoers: ohhh I see you were testing the lerp logic, sorry about that this makes much more sense now. You can add the logic back my bad. Just make sure you're checking against what the truncated lerp result would be i.e. 2.5 --> 2...

Arm/test/joint_test.cpp
```diff
@@ -52,24 +52,45 @@ TEST_CASE("Drive system testing")
...
   {
     joint.SetSpeed(5);
     CHECK(joint.GetSpeed() == 5);
-    joint.SetSpeed(10);
-    CHECK(joint.GetSpeed() == 10);
-    joint.SetSpeed(45);
-    CHECK(joint.GetSpeed() == 45);
-    joint.SetSpeed(100);
-    CHECK(joint.GetSpeed() == 100);
+    joint.SetSpeed(101);
```
naterpotatoers:
```suggestion
    joint.SetSpeed(101);
    CHECK(joint.GetSpeed() == 100);
```

Arm/test/joint_test.cpp
```diff
@@ -52,24 +52,45 @@ TEST_CASE("Drive system testing")
...
+    CHECK(joint.GetSpeed() == 5);
+    joint.SetSpeed(-101);
+    CHECK(joint.GetSpeed() == 5);
+    joint.SetPosition(45);
+    joint.SetPosition(200);
+    CHECK(joint.GetPosition() == 45);
+    joint.SetPosition(-10);
+    CHECK(joint.GetPosition() == 45);
```
naterpotatoers: maybe we set the speed back to zero and check that decreasing lerp works properly. Then inside a new SECTION we test the same logic but for negative angles. Then we would have tested all the [boundaries ](https://www.educba.com/boundary-value-testing/)for SetPosition()

naterpotatoers (approved):
looks good, feel free to merge

Andrew-Ingerman:
image

Is this a problem with the script or is it because some of my tests failed?

naterpotatoers:
> image
> 
> Is this a problem with the script or is it because some of my tests failed?

One of the test cases was broken causing the script to not work right. It is now fixed, when a test case doesn't pass it won't crash just tell you not all test cases passed

PR #89, opened 2022-01-11
Adrien drive refactoring double to floats and esp timer system

acarrou:
In the main file, I instantiated the timer object so we start it each time and then pass it to the esp object member function IsServerExpired(). I believe this to be the correct placement of the function in main but there may be a better place for this. I did attempt to put all of this in GET() but it caused some issues.

### DIFF
--- a/Drive/source/main.cpp
+++ b/Drive/source/main.cpp
@@ -59,16 +59,10 @@ int main(void)
       sjsu::LogInfo("Making new request now...");
       std::string endpoint = "drive" + drive.GETParameters();
       std::string response = esp.GET(endpoint);
-      sjsu::TimeoutTimer serverTimeout(5s);  // server has 5s timeout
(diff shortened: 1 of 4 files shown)


Drive/source/main.cpp
```diff
@@ -56,19 +56,15 @@ int main(void)
...
       drive.ParseJSONResponse(response);
       drive.HandleRoverMovement();
       drive.IncrementHeartbeatCount();
       drive.PrintRoverData();
-      if (serverTimeout.HasExpired())
-      {
-        sjsu::LogWarning("Server timed out! Reconnecting...");
-        esp.ConnectToServer();
-      }
+      esp.IsServerExpired(serverTimeout);
```
naterpotatoers, line 67: ah I see, that's a good way to solve the problem

naterpotatoers (approved):

PR #91, opened 2022-01-12
Fixed homing

Coreyboy1820:
Changed the logic for the homing code slightly


Coreyboy1820:
VS code was acting weird again, hopefully nothing is too different like last time.

naterpotatoers:
There's no code changes?

PR #92, opened 2022-01-12
Constraints for set pitch position and set roll position

Mchan2003:


Arm/test/wrist_joint_test.cpp
```diff
@@ -63,22 +63,64 @@ TEST_CASE("Wrist joint system testing")
...
+  {
+    wrist_joint.SetPitchPosition(-1);
+    CHECK(wrist_joint.GetPitchPosition() == 0);
+  }
+
+  SECTION("3.5 Should return 180 when setting pitch to 181")
+  {
+    wrist_joint.SetPitchPosition(181);
+    CHECK(wrist_joint.GetPitchPosition() == 180);
+  }
```
naterpotatoers, line 93: Since we had unit tests written for going beyond minimum and maximum angles we were able to verify that the logic was working correctly while we were developing. This accidentally showed the power of test driven development where you write unit tests prior to writing out the logic. Congrats :)

naterpotatoers (approved):
feel free to merge in now

PR #94, opened 2022-01-12
 Added Getters setters to min and max angle as well as reset default

ryerlic:
adds diagnostic gets and sets to the joint

### DIFF
--- a/Arm/joint.hpp
+++ b/Arm/joint.hpp
@@ -94,6 +94,37 @@ class Joint
     return int(position_);
   }
 
+
+// edits the min angle. Useful for specific robot modes where full traversal might not be a good idea.
+  void SetMinAngle(float newmin)
+  {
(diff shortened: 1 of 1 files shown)


Arm/joint.hpp
```diff
@@ -94,6 +94,37 @@ class Joint
     return int(position_);
   }
 
+
+// edits the min angle. Useful for specific robot modes where full traversal might not be a good idea.
+  void SetMinAngle(float newmin)
```
naterpotatoers, line 99:
```suggestion
  void SetMinAngle(float new_min)
```
Google style guide naming convention

naterpotatoers (changes requested):
I'm not sure if this logic is entirely necessary, it adds unnecessary bulk. For one I think maximum and minimum angles should be constants that shouldn't change after initialization. Please refer to github issues found on urc-central-2021 for things to work on

PR #96, opened 2022-01-13
Homing wrist

Coreyboy1820:
Homed WristPitch:
  -Had to slightly fill out the hand class to do so.
  -changed a few other things as well
  -couldn't compile because of an error in main that I didn't want to touch.


Coreyboy1820:
I can fix whatever you need me to before this gets merged, it's a large PR.

Arm/rover_arm_system.hpp
```diff
@@ -67,8 +67,9 @@ class RoverArmSystem : public sjsu::common::RoverSystem
   RoverArmSystem(sjsu::arm::Joint & rotunda,
                  sjsu::arm::Joint & shoulder,
                  sjsu::arm::Joint & elbow,
+                 sjsu::arm::Hand & hand,
                  sjsu::arm::WristJoint & wrist)
-      : rotunda_(rotunda), shoulder_(shoulder), elbow_(elbow), wrist_(wrist)
+      : rotunda_(rotunda), shoulder_(shoulder), elbow_(elbow), hand_(hand), wrist_(wrist)
```
naterpotatoers: This is why main.cpp has an error, it needs to reflect the new constructor which now has a hand argument

naterpotatoers (changes requested):
Fix the initialization in main.cpp and should be good to merge in

naterpotatoers (approved):

PR #97, opened 2022-01-13
Created structs to use for cleaner error handling

acarrou:
Although having structs to throw exceptions in our code for errors. I think it makes main.cpp looks cluttered. Maybe there's a better way to do this. I think this is something we can discuss in the meeting at some point.


naterpotatoers (changes requested):
Either fix the unit test or comment out the failing tests momentarily until they get fixed

naterpotatoers (approved):
unit tests are failing becaues IsHeartbeatSynced() now takes an argument

PR #100, opened 2022-01-14
Cleaning up/unit fixing

Coreyboy1820:
Cleaned up units:
 -switched mission control structure data from floats to ints.
 -switched the degrees and rpm in the joint class constructor/member variables to floats.
 -as a result, also had to switch the main file and test for arm system.
GetPosition() fix:
 -switched the angle - zero offset to addition.


naterpotatoers (approved):
looks good

PR #102, opened 2022-01-14
Heartbeat base class tests

naterpotatoers:
adds tests to common dir with testing for heartbeat logic. removes duplication from arm and drive.
removes heartbeat unit tests from drive


acarrou:
looks good to me, nice job!

acarrou (approved):

PR #103, opened 2022-01-14
Creates function to get encoder data for wheel homing

acarrou:
We will need to create const variables for each wheel's home position using the encoder position in the RmdX class.


naterpotatoers (approved):
looks good

PR #107, opened 2022-01-15
Lerp for arm

Andrew-Ingerman:


Arm/joint.hpp
```diff
@@ -63,11 +63,9 @@ class Joint
     return acceleration;
   }
 
-  void SetJointSpeed(float target_speed)
-  {
-    speed_ = std::lerp(speed_, target_speed, kLerpStep);
-    units::angular_velocity::revolutions_per_minute_t speed_to_rpm(speed_);
-    motor_.SetSpeed(speed_to_rpm);
+  void SetJointSpeed(double targetspeed){
```
naterpotatoers, line 66:
```suggestion
  void SetJointSpeed(float target_speed){
```
We should keep everything in type float for consistency.

naterpotatoers (changes requested):
I think your branch is a bit out of date, try updating your branches then redoing the logic to prevent overwriting prior updates

PR #108, opened 2022-01-15
arm data printout

ulysses10:
adds arm data print

### DIFF
--- a/Arm/rover_arm_system.hpp
+++ b/Arm/rover_arm_system.hpp
@@ -85,7 +85,42 @@ class RoverArmSystem : public sjsu::common::RoverSystem
 
   void PrintRoverData() override
   {
-    return;
+    printf("Arm data: \n");
+    printf("Mode: %c\n", mc_data_.modes);
+    printf("Arm speed: %d\n", mc_data_.arm_speed);
+    printf("Mode: %d\n", mc_data_.rotunda_angle);
+    printf("Mode: %d\n", mc_data_.shoulder_angle);
+    printf("Mode: %d\n", mc_data_.elbow_angle);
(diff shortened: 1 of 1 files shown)


Arm/rover_arm_system.hpp
```diff
@@ -85,7 +85,42 @@ class RoverArmSystem : public sjsu::common::RoverSystem
...
   {
-    return;
+    printf("Arm data: \n");
+    printf("Mode: %c\n", mc_data_.modes);
+    printf("Arm speed: %d\n", mc_data_.arm_speed);
+    printf("Mode: %d\n", mc_data_.rotunda_angle);
+    printf("Mode: %d\n", mc_data_.shoulder_angle);
+    printf("Mode: %d\n", mc_data_.elbow_angle);
+    printf("Mode: %d\n", mc_data_.wrist_roll);
+    printf("Mode: %d\n", mc_data_.wrist_pitch);
```
naterpotatoers:
```suggestion
    printf("Rotunda Angle: %d\n", mc_data_.rotunda_angle);
    printf("Shoulder Angle: %d\n", mc_data_.shoulder_angle);
    printf("Elbow Angle: %d\n", mc_data_.elbow_angle);
    printf("Wrist Roll Angle: %d\n", mc_data_.wrist_roll);
    printf("Wrist Pitch Angle: %d\n", mc_data_.wrist_pitch);
```

naterpotatoers (changes requested):

naterpotatoers (approved):
looks good

joshua-git0:
We could also add another function to print this out in greater significant digits (like float, double, or long double) and it might be useful to get current or voltage draw when the arm is being set to an angle/speed @naterpotatoers @ulysses10

PR #109, opened 2022-01-15
Encoder position

acarrou:
just simple encoder position testing.

### DIFF
--- a/Drive/rover_drive_system.hpp
+++ b/Drive/rover_drive_system.hpp
@@ -228,6 +228,7 @@ class RoverDriveSystem : public sjsu::common::RoverSystem
     printf("DRIVE MODE:\t%d\n", current_mode_);
     printf("MC SPEED:\t%d\n", mc_data_.speed);
     printf("MC ANGLE:\t%d\n", mc_data_.rotation_angle);
+    printf("Encoder Position:\t%d\n", right_wheel_.GetHubEncoderPos());
     printf("WHEEL     SPEED     ANGLE\n");
     printf("=========================\n");
     left_wheel_.Print();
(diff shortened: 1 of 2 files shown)


Drive/rover_drive_system.hpp
```diff
@@ -228,6 +228,7 @@ class RoverDriveSystem : public sjsu::common::RoverSystem
     printf("DRIVE MODE:\t%d\n", current_mode_);
     printf("MC SPEED:\t%d\n", mc_data_.speed);
     printf("MC ANGLE:\t%d\n", mc_data_.rotation_angle);
+    printf("Encoder Position:\t%d\n", right_wheel_.GetSteerEncoderPos());
```
naterpotatoers:
```suggestion
```
maybe we should move this to the Wheel class Print() function

Drive/wheel.hpp
```diff
@@ -67,9 +67,9 @@ class Wheel
     return homing_offset_angle_;
   }
 
-  int GetEncoderPos(sjsu::RmdX motor)
+int GetSteerEncoderPos()
```
naterpotatoers:
```suggestion
int GetSteerEncoderPosition()
```

naterpotatoers (changes requested):
I think we should just print out all the encoder positions for left, right, and back wheels since we'll need to do that at some point

acarrou:
Still need to edit this. Will complete it tomorrow

naterpotatoers (approved):
talked with Khalil about using encoder position apparently this solution would not work as it's not truly a fixed position. He spoke with manufacturer a ton prior to buying these and wanted clarification whether we could home using just the device by itself but that is not the case. With that said we can still see what values this prints out but also we should look into homing using mission control. Maybe we could home it with mission control making small adjustments on each wheel somehow i don't know how complicated that would be but it sounds like we might be able to repurpose the singular wheel modes for this in the meantime.

acarrou:
> talked with Khalil about using encoder position apparently this solution would not work as it's not truly a fixed position. He spoke with manufacturer a ton prior to buying these and wanted clarification whether we could home using just the device by itself but that is not the case. With that said we can still see what values this prints out but also we should look into homing using mission control. Maybe we could home it with mission control making small adjustments on each wheel somehow i don't know how complicated that would be but it sounds like we might be able to repurpose the singular wheel modes for this in the meantime.

Ah darn, that's annoying. Usually, with these motors, the encoder position is super helpful :/ I can try and come up with another solution. Thanks for letting me know @naterpotatoers

PR #111, opened 2022-01-15
Add enum class for different drive modes

joshua-git0:
@naterpotatoers Is this what you wanted? All I did was place an enum class in the private variables in **class RoverDriveSystem : public sjsu::common::RoverSystem**


Drive/rover_drive_system.hpp
```diff
@@ -237,6 +237,20 @@ class RoverDriveSystem : public sjsu::common::RoverSystem
...
+  // https://github.com/SJSURoboticsTeam/urc-central-2021/issues/196
+  // S, D, T, L, R, B = Spin, Drive, Translate, Left, Right, Back.
+  enum class Modes : char
+    {
+      kDefault = 'D',
+      SpinMode = 'S'
+      TranslateMode = 'T'
+      LeftMode = 'L'
+      RightMode = 'R'
+      BackMode = 'B'
```
naterpotatoers:
```suggestion
      DriveMode= 'D',
      SpinMode = 'S'
      TranslateMode = 'T'
      LeftWheelMode = 'L'
      RightWheelMode = 'R'
      BackWheelMode = 'B'
```

Drive/rover_drive_system.hpp
```diff
@@ -237,6 +237,20 @@ class RoverDriveSystem : public sjsu::common::RoverSystem
   }
 
  private:
+
+  // https://github.com/SJSURoboticsTeam/urc-central-2021/issues/196
```
naterpotatoers:
```suggestion
```

Drive/rover_drive_system.hpp
```diff
@@ -237,6 +237,20 @@ class RoverDriveSystem : public sjsu::common::RoverSystem
...
+  // S, D, T, L, R, B = Spin, Drive, Translate, Left, Right, Back.
```
naterpotatoers:
```suggestion
```
We should no longer need to have additional documentation on what the drive modes are as they clearly define themselves now. This is a big improvement to the code quality because needing to document what your code is doing is a sign of bad design as good code should read like good prose and be fully explicit.

Drive/rover_drive_system.hpp
naterpotatoers: You can mention the issue in the pull request - not in the code itself as it clutters the code base and makes comments less important

naterpotatoers (changes requested):
good start, this is exactly what we are looking for.

joshua-git0:
Thanks @naterpotatoers  for the comments! Sorry about that, I usually add comments as notes to myself but for sure I should've deleted it for this pull request

naterpotatoers:
any updates on this?

Drive/rover_drive_system.hpp
```diff
@@ -237,6 +237,18 @@ class RoverDriveSystem : public sjsu::common::RoverSystem
...
+  enum class Modes : char
+    {
+      DriveMode= 'D',
+      SpinMode = 'S'
+      TranslateMode = 'T'
+      LeftWheelMode = 'L'
+      RightWheelMode = 'R'
+      BackWheelMode = 'B'
+      
+    };
```
naterpotatoers: This doesn't compile. I think you are missing the `,`'s

naterpotatoers (changes requested):

acarrou:
Will Need to update unit tests but the code works after testing it on sjtwo and server.

Drive/rover_drive_system.hpp
```diff
@@ -35,7 +35,6 @@ class RoverDriveSystem : public sjsu::common::RoverSystem
   struct MissionControlData : public RoverMissionControlData
   {
     int wheel_shift    = 0;
-    char drive_mode    = 'S';
```
naterpotatoers, line 38: We should not be removing this from MissionControlData. It is important for detecting a change in drive mode.

Drive/rover_drive_system.hpp
```diff
@@ -382,32 +381,32 @@ class RoverDriveSystem : public sjsu::common::RoverSystem
...
   int state_of_charge_ = 90;
-  char current_mode_   = 'S';
 
   const int kExpectedArguments = 6;
   const float kZeroSpeed       = 0;
   const float kMaxTurnRadius   = 45;
   const float kLerpStep        = 0.5;
 
  public:
+ Modes current_drive_mode_ = Modes::SpinMode;
```
naterpotatoers: I do like the update member variable name, I think it is more explanatory.

Drive/rover_drive_system.hpp
```diff
@@ -244,27 +243,26 @@ class RoverDriveSystem : public sjsu::common::RoverSystem
...
+      RightWheelMode = 'R',
       BackWheelMode = 'B'
-      
     };
 
   /// Stops the rover and sets a new mode.
   void SetMode()
   {
-    sjsu::LogWarning("Switching rover into %c mode...", mc_data_.drive_mode);
+    sjsu::LogWarning("Switching rover into %c mode...", current_drive_mode_);
```
naterpotatoers: this should still be referencing the new mc data drive mode

Drive/rover_drive_system.hpp
```diff
@@ -244,27 +243,26 @@ class RoverDriveSystem : public sjsu::common::RoverSystem
...
     StopWheels();
-    switch (mc_data_.drive_mode)
+    switch (current_drive_mode_)
```
naterpotatoers: stick to using mc data drive mode since current_drive_mode_ isn't updated until calling one of the Set(mode)Handlers()

Drive/rover_drive_system.hpp
```diff
@@ -141,7 +140,7 @@ class RoverDriveSystem : public sjsu::common::RoverSystem
   /// Checks if the rover got a new drive mode command
   bool IsNewMode()
   {
-    if (current_mode_ != mc_data_.drive_mode)
+    if (current_drive_mode_ != current_drive_mode_)
```
naterpotatoers: This is a major bug!!! This is why we needed to have a mc drive mode variable to compare it to our current drive mode

Drive/rover_drive_system.hpp
```diff
@@ -79,7 +78,7 @@ class RoverDriveSystem : public sjsu::common::RoverSystem
     int actual_arguments =
         sscanf(response.c_str(), response_body_format,
                &mc_data_.heartbeat_count, &mc_data_.is_operational, &mc_data_.wheel_shift,
-               &mc_data_.drive_mode, &mc_data_.speed, &mc_data_.rotation_angle);
+               &current_drive_mode_, &mc_data_.speed, &mc_data_.rotation_angle);
```
naterpotatoers: major bug!! we need to keep track of current drive mode to compare it to the new mc data drive mode. Directly overwriting the current drive mode makes this impossible

naterpotatoers (changes requested):
Major bug found, needs fix. Good work otherwise

Drive/rover_drive_system.hpp
acarrou: yea forgot about this

Drive/rover_drive_system.hpp
acarrou: This would mean I put the enum into the mission control struct in order to use the drive_mode variable

naterpotatoers: Does it have to be? Can we not reuse the Modes type? I remember this being problematic in the past which is why we stuck with basic char type. Try declaring the Modes struct before the MissionControlData struct ?

acarrou: I just put the enum into the Mission Control data. It works fine but wasn't sure if that's what you wanted

naterpotatoers: hmm normally i would say its fine but I think it might be best if we left it as it's own struct within public. Honestly I think it might be a good idea to move a lot of these structs to their own files. Cause now we have so many structs that we are having to recreate in different locations versus just importing the struct and just using it like a normal variable like char

acarrou: Yea that's probably a good idea. The issue is structs and chars cannot be compared or converted since they are not of the same type. Maybe I'll spend  more time on this so that it looks a lot cleaner and like you said use them in another file

naterpotatoers: yeah this might be a bit tricky.

naterpotatoers (approved):

acarrou:
sorry for so many commits

PR #112, opened 2022-01-16
Finger class

Coreyboy1820:
Started finger class:
-Created constructor/changed hand constructor.
-Created member variables.
-Created the initialize function for finger and hand.
-Changed the hand class to construct the fingers.
-Changed main to account for new finger class.


naterpotatoers (approved):
looks good merge in when ready

PR #113, opened 2022-01-16
Hand setters getter

Mchan2003:


Arm/Hand/hand.hpp
```diff
@@ -27,52 +27,101 @@ class Hand
 
   Hand(sjsu::arm::WristJoint wrist) : wrist_(wrist) {}
```
naterpotatoers:
```suggestion
  Hand(sjsu::arm::WristJoint & wrist) : wrist_(wrist) {}
```
Try this fix and it should fix errors

naterpotatoers (approved):

naterpotatoers:
Do not merge until this compiles*** I just approved it out of courtesy since it should be an easy fix

Arm/Hand/hand.hpp
Mchan2003: Hi, I tried to compile the code with your fix but it keeps giving me this error: 

` error: uninitialized reference member in 'class sjsu::RmdX&' [-fpermissive]
   28 | Hand(sjsu::arm::WristJoint & wrist) : wrist_(wrist) {}
   |      ^~~~`

Is the error referring to Hand being uninitialized or is there a reference member in the RmdX class being uninitialized?

Mchan2003: Nevermind. I think I solved the error.

naterpotatoers: sorry didn't see this earlier

Arm/Hand/hand.hpp
```diff
@@ -8,15 +8,15 @@ class Hand
...
-    float pinky_angle   = 0;
-    float ring_angle    = 0;
-    float middle_angle  = 0;
-    float pointer_angle = 0;
-    float thumb_angle   = 0;
+    float pinky_angle_   = 0;
+    float ring_angle_    = 0;
+    float middle_angle_  = 0;
+    float pointer_angle_ = 0;
+    float thumb_angle_   = 0;
```
naterpotatoers: Leave the trailing underscore off. Here's the [Google style guide reference](https://google.github.io/styleguide/cppguide.html#Variable_Names)

Arm/Hand/hand.hpp
naterpotatoers: We need to pass by reference here still 
```suggestion
  Hand(sjsu::arm::WristJoint & wrist) : wrist_(wrist) {}
```

Arm/Hand/hand.hpp
```diff
@@ -27,52 +27,101 @@ class Hand
...
+    units::angle::degree_t angle_to_degrees(target_pointer_angle);
+    pointer_motor_.SetAngle(angle_to_degrees);
+  } ;
+
+  void SetRingPosition(float target_ring_angle)
+  {
+    fingers_.ring_angle_ = float(std::clamp(target_ring_angle, min_angle, max_angle));
+    units::angle::degree_t angle_to_degrees(target_ring_angle);
+    ring_motor_.SetAngle(angle_to_degrees);
+  };
```
naterpotatoers, line 102: Since we've implemented the Finger.hpp class these actually will go away and be handled in the Finger.hpp class now

Arm/Hand/hand.hpp
```diff
@@ -27,52 +27,101 @@ class Hand
...
   {
-    return 0;
+    return int(fingers_.pointer_angle_);
   };
 
   int GetRingPosition()
   {
-    return 0;
+    return int(fingers_.ring_angle_);
   };
```
naterpotatoers, line 128: As mentioned before, these will be replaced by the finger.hpp class which should be public member variables of this class

naterpotatoers (changes requested):
Honestly with the addition of the finger.hpp class we can probably just close out this PR without merging it since we will instead be calling the Finger.hpp class within this class for moving the fingers etc. Sorry about that, probably should have spent more time brainstorming before implementing

PR #114, opened 2022-01-17
Unit test clean up

Coreyboy1820:
Cleaned up unit testing for both drive and arm:
-Erased unneeded scope resolution.
-Took away unneeded declarations of mocks.
-However, not sure why we need all of the functions of the spys, some clarification would be nice.


Drive/test/rover_drive_system_test.cpp
```diff
@@ -20,12 +20,7 @@ TEST_CASE("Drive system testing")
   StaticMemoryResource<1024> memory_resource;
   CanNetwork network(mock_can.get(), &memory_resource);
 
-  RmdX left_steer_motor(network, 0x141);
-  RmdX left_hub_motor(network, 0x142);
-  RmdX right_steer_motor(network, 0x143);
-  RmdX right_hub_motor(network, 0x144);
-  RmdX back_steer_motor(network, 0x145);
-  RmdX back_hub_motor(network, 0x146);
+  RmdX motor(network, 0x141);
```
naterpotatoers, line 23: Did you verify this actually works as anticipated?

naterpotatoers (approved):

Drive/test/rover_drive_system_test.cpp
Coreyboy1820: Honestly no, I thought the fact that all the tests passed was good enough, how would I go about verifying this? I'm not too sure how I would do this.

PR #116, opened 2022-01-18
C kelley pitch offset fix

Coreyboy1820:
Fixed up the pitch offset:
-Offset had to be added to the pitch/roll to allow for accurate moving
Put a lot of to do's in code for clean up.
Added more drive modes for arm.


naterpotatoers (approved):
Looks good

PR #117, opened 2022-01-19
added logic to getters and setters for position and speed

Mchan2003:


naterpotatoers (approved):
good start, not sure if these servos have a speed component to them though but I guess we can test it and find out later.

PR #118, opened 2022-01-20
Arm system clean up

Coreyboy1820:
Cleaned up Arm System:
-Broke down homing functions to more descriptive functions.
-Moved homing to private.
-No need for the accelerations structure in the arm system class, removed.
-Added acceleration member variables in wrist and joint classes.
-Fixed unit testing with the new implementation of CheckIfZero affecting the mpu values from GetAccelerometerData.


Arm/Hand/wrist_joint.hpp
```diff
@@ -93,10 +93,19 @@ class WristJoint
...
  private:
 
+  /// Checks if value is zero. If it's zero make it not zero
+  float ReturnChangedIfZero(float acceleration)
+  {
+    if (acceleration == 0)
+    {
+      acceleration = 0.001;
+    }
+    return acceleration;
```
naterpotatoers, line 105: could 1 line it using ternary operator like so:
`return acceleration == 0 ? 0.001 : acceleration;`

Arm/Hand/wrist_joint.hpp
```diff
@@ -63,14 +63,14 @@ class WristJoint
     roll_offset_angle_ = roll_offset;
   }
 
-  /// Return the acceleration values for the MPU6050 on the joint.
+  //return is only used for testing
```
naterpotatoers, line 66: If the return value is only used for testing and the acceleration_ member variable is public, then it seems redundant. You could make this function void or make acceleration_ private. imo keeping acceleration_ public is easier  since we can potentially use that to directly modify the acceleration values for other testing purposes

Arm/joint.hpp
```diff
@@ -89,6 +83,16 @@ class Joint
...
  private:
+ 
+  float ReturnChangedIfZero(float acceleration)
+  {
+    if (acceleration == 0)
+    {
+      acceleration = 0.001;
+    }
+    return acceleration;
+  }
```
naterpotatoers, line 94: I notice we have a few repeated functions in both joint.hpp and wrist_joint.hpp. Might be worth creating a base class in the future to remove duplicated logic.

naterpotatoers: maybe we make joint.hpp be the base and make arm joint logic be arm_joint.hpp?

Arm/rover_arm_system.hpp
```diff
@@ -194,68 +186,100 @@ class RoverArmSystem : public sjsu::common::RoverSystem
...
+
+  bool InThirdQuandrantOfGraph()
+  {
+  if (elbow_.acceleration_.x + rotunda_.acceleration_.x >= 0 &&
+             elbow_.acceleration_.y + rotunda_.acceleration_.y >= 0)
+      {
+        return true;
+      }
+  return false;
+  }
```
naterpotatoers, line 262: yesssssss

Arm/rover_arm_system.hpp
```diff
@@ -194,68 +186,100 @@ class RoverArmSystem : public sjsu::common::RoverSystem
...
+  float CalculateUncorrectedElbowHomeAngle()
+  {
+    float acceleration_x = rotunda_.acceleration_.x + elbow_.acceleration_.x;
+    float acceleration_y = rotunda_.acceleration_.y + elbow_.acceleration_.y;
+    float angle_without_correction =
+        float(atan(acceleration_y / acceleration_x));
+    return angle_without_correction;
+  }
+
+  bool InSecondQuadrantOfGraph()
```
naterpotatoers, line 244:
```suggestion
  bool IsInSecondQuadrantOfGraph()
```
Maybe rename these to be more descriptive? What exactly is in second quadrant?

naterpotatoers (approved):
looks good. the suggestions can be added in later PRs

Arm/joint.hpp
Coreyboy1820: I was starting to think the same thing lol, will do.

Arm/Hand/wrist_joint.hpp
Coreyboy1820: we use the acceleration_ in the rover arm system, so I'll keep it public and make it a void function.

PR #120, opened 2022-01-20
Joint base class

Coreyboy1820:
Created the joint base class for the wrist joint and arm joint:
-Changed the joint to armjoint everywhere.
-Moved all similar code to the joint base class
-Moved Acceleration Structure.
-Fixed unit testing for the new base class.
Tested to make sure that the current motor setup in unit testing would work and it does:
-Rotunda.GetPosition() returns a different value then elbow.GetPosition() when they use the same motor
-Tested by setting both positions first then comparing them using CHECK_FALSE.


Arm/project.mk
```diff
@@ -1,5 +1,5 @@
-TESTS += test/wrist_joint_test.cpp
+#TESTS += test/wrist_joint_test.cpp
 TESTS += test/rover_arm_system_test.cpp
-TESTS += test/joint_test.cpp
-TESTS += test/finger_test.cpp
-TESTS += test/hand_test.cpp
\ No newline at end of file
+#TESTS += test/arm_joint_test.cpp
+#TESTS += test/finger_test.cpp
+#TESTS += test/hand_test.cpp
```
naterpotatoers: uncomment test cases

naterpotatoers (approved):

PR #123, opened 2022-01-22
Implemented HandleHandMovement()

Mchan2003:


naterpotatoers (approved):
nicely done. just need unit tests which can be done at a later point

PR #126, opened 2022-01-25
created unit tests for finger

Andrew-Ingerman:
What's the difference between CHECK and CHECK_EQ

### DIFF
--- a/Arm/test/finger_test.cpp
+++ b/Arm/test/finger_test.cpp
@@ -6,13 +6,39 @@ namespace sjsu::arm
 {
 TEST_CASE("Finger Testing Case...")
 {
-    Mock<Servo> mock_servo;
+  Mock<Servo> mock_servo;
 
-    Finger finger(mock_servo.get());
+  Finger finger(mock_servo.get());
 
-    SECTION("Should return 0 when called...")
-    {
-        CHECK_EQ(finger.GetPosition(), 0);
-    }
+  SECTION("Should return 0 when finger is first initialized")
+  {
+    CHECK_EQ(finger.GetPosition(), 0);
+  }
+
+  SECTION("Should return 0, 10, 45, 90, 100 when SetPosition() is called")
+  {
+    CHECK_EQ(finger.GetPosition(), 0);
+    finger.SetPosition(10);
+    CHECK_EQ(finger.GetPosition(), 10);
+    finger.SetPosition(45);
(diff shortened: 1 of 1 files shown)


Arm/test/finger_test.cpp
```diff
@@ -6,13 +6,39 @@ namespace sjsu::arm
...
+  SECTION("Boundry testing the angles")
+  {
+    int max_angle_ = 180;
+    int min_angle_ = 0;
+    CHECK_EQ(finger.GetPosition(), 0);
+    finger.SetPosition(-1);
+    CHECK_EQ(finger.GetPosition(), 0);
+    finger.SetPosition(181);
+    CHECK_EQ(finger.GetPosition(), 180);
+  }
```
naterpotatoers: Refer to [testing document](https://docs.google.com/document/d/1GoI3ypr8xW_N2GrNbO3eNoYoz85498nix4rTv1WexoA/edit) e.g.
```c++
SECTION("2.1 Should set position to 0 when set to 0")
{
// Unit test...
}

SECTION("2.2 Should set position to 0 when set to -1")
{
// Unit test...
}
```
This will help us keep track of the unit tests and what is already tested and what needs to be tested. Otherwise it is easy to duplicate tests without realizing it

naterpotatoers (changes requested):
CHECK() checks for trueness. So as long as the condition inside returns true it will pass the test case e.g. bool return type is true or random_num == expected_num etc.
CHECK_EQ() checks for equality. This just verifies that the two arguments are equal whether that be two ints or two strings e.g. "word" == "word" or 1 == 1. You could use CHECK() instead of CHECK_EQ() but its just preference I suppose. I think it looks cleaner

Arm/test/finger_test.cpp
```diff
@@ -6,13 +6,24 @@ namespace sjsu::arm
...
+  SECTION("2.1 - 2.4 Boundry testing the angles")
+  {
+    int max_angle_ = 180;
+    int min_angle_ = 0;
+    CHECK_EQ(finger.GetPosition(), 0);
+    finger.SetPosition(-1);
+    CHECK_EQ(finger.GetPosition(), 0);
+    finger.SetPosition(181);
+    CHECK_EQ(finger.GetPosition(), 180);
+  }
```
naterpotatoers: Generally it's good practice to separate the test cases in case one fails. That way it'll be easier to identify which case is causing problems

PR #127, opened 2022-01-25
Switch case for arm modes

Coreyboy1820:
Created drive modes for arm and hand:
-Implemented with switch cases.
-Created a new enum of modes for the hand in the mc data.
-different modes consist of concurrent for single joint and full arm system, each joint individually, an open and close mode for the hand, a mode for the pitch and roll as of right now, and a mode for homing the arm and then the hand.
-Implemented a few helper functions.


Arm/rover_arm_system.hpp
```diff
@@ -34,18 +34,26 @@ class RoverArmSystem : public sjsu::common::RoverSystem
...
+    };
+    ArmModes ArmMode = ArmModes::kConcurrent;
+    enum class HandModes : char
+    {
+      kPitch      = 'P',
+      kRoll       = 'R',
+      kClose      = 'F',
+      kOpen       = 'O',
+      kConcurrent = 'C'
     };
```
naterpotatoers: Update [POR ](https://docs.google.com/presentation/d/17clPo-6g9CDImaTh3-akEh0AcbR9gbY4OoO5jGr4u0E/edit#slide=id.p)document (after slide ~76) with all these new modes. That way we can better understand each of these modes' purpose

Arm/rover_arm_system.hpp
```diff
@@ -34,18 +34,26 @@ class RoverArmSystem : public sjsu::common::RoverSystem
...
       kHomeHand   = 'H',
       kConcurrent = 'C',
       kRotunda    = 'R',
       kShoulder   = 'S',
       kElbow      = 'E',
-      kWrist      = 'W'
-
+      kHand       = 'D'
+    };
+    ArmModes ArmMode = ArmModes::kConcurrent;
```
naterpotatoers:
```suggestion
    ArmModes ArmMode = ArmModes::kConcurrent;
```
move this line down below with the others, almost didn't realize it was still here.

naterpotatoers (approved):
nicely done

Arm/rover_arm_system.hpp
Coreyboy1820: got it

Arm/rover_arm_system.hpp
Coreyboy1820: Done.

PR #131, opened 2022-02-04
Seeding wrist class in hand class

Coreyboy1820:
Seeded wrist inside of hand:
-Removed dependencies of rover arm system on wrist
-Created functions inside of the hand class to allow for use of the wrists functions since inheritance wasn't used to get these functions.
-Fixed unit testing.
-Deleted unnecessary acceleration struct in hand.
-Added speed to the wrist joint functions to actually make the wrist move.
     -Added new member variable for speed in wrist along with min/max variables as well.
-Made the hand/wrist side of the rover_arm_system more organized, i.e. hand functions like home inside of hand, etc.


Arm/Hand/hand.hpp
```diff
@@ -41,17 +31,64 @@ class Hand
...
+
+  void HandleHandMovement(
+      float speed,
+      float thumb_position,
+      float pointer_position,
+      float middle_position,
+      float ring_position,
+      float pinky_position, 
+      float roll_position, 
+      float pitch_position)
```
naterpotatoers, line 55: At some point we should convert this to a struct since having a ton of parameters is bad practice

Arm/joint.hpp
```diff
@@ -6,32 +6,39 @@ namespace sjsu::arm
...
+  {
     float x = 0;
     float y = 0;
     float z = 0;
- };
+  };
 
-Joint(sjsu::Mpu6050 & accelerometer) : mpu_(accelerometer)
-{};
+  Joint(sjsu::Mpu6050 & accelerometer) : mpu_(accelerometer){};
```
naterpotatoers:
```suggestion
  Joint(sjsu::Accelerometer & accelerometer) : mpu_(accelerometer){};
```

Arm/joint.hpp
```diff
@@ -6,32 +6,39 @@ namespace sjsu::arm
...
 
   /// Checks if value is zero. If it's zero make it not zero
   float ReturnChangedIfZero(float acceleration)
   {
     return (acceleration == 0 ? .001 : acceleration);
   }
 
-sjsu::Mpu6050 & mpu_;
-Acceleration acceleration_;
+  sjsu::Mpu6050 & mpu_;
```
naterpotatoers:
```suggestion
  sjsu::Accelerometer & accelerometer_;
```

Arm/rover_arm_system.hpp
```diff
@@ -191,42 +187,96 @@ class RoverArmSystem : public sjsu::common::RoverSystem
                 float pointer,
                 float middle,
                 float ring,
-                float pinky)
+                float pinky,
+                float pitch,
+                float roll)
```
naterpotatoers, line 213: Same as mentioned earlier, will need to create a struct to make this cleaner and best practice

naterpotatoers (approved):
This is an important update, good work

Arm/Hand/hand.hpp
Coreyboy1820: I wanted to do this in a different PR, but will do.

PR #133, opened 2022-02-05
start of arm subclass

Coreyboy1820:
Started arm sub-class:
-made header file
-started moving functions from rover arm system to arm.


Arm-Subsytem/Arm/human_arm.hpp
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  void HandleConcurrentMode(float rotunda_angle,
+                            float shoulder_angle,
+                            float elbow_angle)
+  {
+    MoveRotunda(rotunda_angle);
+    MoveShoulder(shoulder_angle);
+    MoveElbow(elbow_angle);
+  }
```
naterpotatoers: I think these should be private

Arm-Subsytem/rover_arm_system.hpp
```diff
@@ -216,10 +190,9 @@ class RoverArmSystem : public sjsu::common::RoverSystem
...
   }
 
-  void HomeArm()
+  void HomeArmSystem()
   {
-    HomeShoulder();
-    HomeElbow();
+    // HomeArm();
     hand_.HomeHand(mc_data_.arm_speed, rotunda_.GetOffsetAngle());
   }
```
naterpotatoers: Need to call the arm_.HomeShoulder() & arm_.HomeElbow functions here most likely

naterpotatoers (approved):
A few small things I noticed that are probably missing should be added before merging.
Nice start for now

naterpotatoers (approved):
looks good

PR #134, opened 2022-02-07
moved even more things that pertained to hand into hand

Coreyboy1820:
Almost done cleaning up the classes. Just need to use the structures inside the mc data in the main functions now.


naterpotatoers (approved):
lets merge this in and we can resolve broken code later

PR #135, opened 2022-02-07
Update unit test for drive private variables

Maaz0070:


naterpotatoers (changes requested):
I think these changes are already merged into master?

Maaz0070:
>
The constructor for the Drive System file was changed; we added a struct for the wheels; and updated unit tests to work with the struct passing. There are changes to this PR that weren't in the other one.

naterpotatoers (approved):
okay looks good then

PR #136, opened 2022-02-08
Switching to struct passing

Coreyboy1820:
Changed all large parameter lists in functions to structures instead:
-In the arm class.
-In the hand class.


Arm-Subsytem/Hand/hand.hpp
```diff
@@ -71,40 +73,53 @@ class Hand
...
 
-  void SetWristPitchPosition(float speed, float pitch_position)
+  void SetWristPitchPosition(float wrist_pitch, float speed)
   {
-    wrist_.SetRollPosition(speed, pitch_position);
+    wrist_.SetRollPosition(wrist_pitch, speed);
+  }
+
+  void HandleMovement(
+                      MissionControlData hand_data, float speed)
```
naterpotatoers:
```suggestion
  void HandleMovement(MissionControlData hand_data, float speed)
```
Run the formatter again on this file

naterpotatoers (approved):
looks good

Arm-Subsytem/Hand/hand.hpp
Coreyboy1820: oh yah I was going to do that, just slipped my mind thanks.

PR #138, opened 2022-02-11
Added const to all getters in arm

Coreyboy1820:
-Added const to all of the getters in the arm classes


naterpotatoers (approved):
looks good

PR #159, opened 2022-03-26
reversed a couple of Gear Ratios

Andrew-Ingerman:


Andrew-Ingerman:
Turns out the ratio's were correct

PR #165, opened 2022-04-23
Stateof charge interface

Andrew-Ingerman:
I expanded the Interface with 3 methods. I then split the logic of the StateOfCharge into three methods CalcState, IsBatteryLow, GetStateOfCharge. I re-implemented the variables including turning some into member and others into local. It seems to work.

### DIFF
--- a/Common/state_of_charge.hpp
+++ b/Common/state_of_charge.hpp
@@ -28,10 +28,10 @@ class StateOfCharge
         new Max170343(i2c, alert_pin, callback, address);
 
     units::voltage::volt_t battery_voltage = battery_sensor->GetVoltage();
-    double voltage                         = battery_voltage.to<double>();
+    voltage                         = battery_voltage.to<double>();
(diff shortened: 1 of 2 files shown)


Common/state_of_charge.hpp
```diff
@@ -12,61 +12,59 @@ class StateOfCharge
 {
  public:
   StateOfCharge() {}
+  sjsu::lpc40xx::I2c & i2c        = sjsu::lpc40xx::GetI2c<2>();
+  sjsu::lpc40xx::Gpio & alert_pin = sjsu::lpc40xx::GetGpio<0, 6>();
+  InterruptCallback callback;
+  uint8_t address;
+  units::voltage::volt_t battery_voltage;
+  float voltage;
```
naterpotatoers, line 18: Move these into private at the bottom of the class

naterpotatoers (approved):

